PR #6, opened 2020-10-17
HM-70 build fix +HM-85

vperuzzi:
This PR is to rectify the failed builds of the previously merged PR HM-70: Landing Page (this occurred as a result of the CI integration occurring after I initially made the PR, I believe). The build was failing due to the removal of the SiteTitle.vue component being removed, but I had failed to remove/alter the corresponding unit test for this component. The detailed build of this PR can be viewed here: https://travis-ci.com/github/steven-murray/TheHaloMod-SPA/jobs/401238653 

This branch resolves these issues by:
1. Factoring out the md-toolbar from the App.vue into it's own component: **Navbar.vue** (for decoupling purposes)
2. Removing the SiteTitle.vue unit test to be a Navbar.vue unit test

Both of the above changes are being tracked in the storyless task, **[HM-85](https://the-halo-mod.atlassian.net/browse/HM-85)** on the Jira board.

Proof of successful build: https://travis-ci.com/github/steven-murray/TheHaloMod-SPA/builds/190681530

@aschnapp I am requesting you as a reviewer because you initially contacted me about the build failure, and you are our CI master 🚀 👑


aneuhold:
@vperuzzi it looks like this PR has some conflicts after the latest merge. That is probably because the latest PR has a fix to the tests so that it tests the About page.

aneuhold:
@vperuzzi never-mind! Fixed it :)

client/src/views/Navbar.vue
```diff
@@ -0,0 +1,20 @@
...
+
+<script>
+// @ is an alias to /src
+
+export default {
+  name: 'Navbar',
+  components: {
+  },
+};
+</script>
```
aneuhold, line 20: Should this be in a `components` folder instead of `views`? It seems like views is for routes. Not a big deal either way. Just might help with organization as we move along.

aneuhold (changes requested):
Just one comment to be looked at really quick. Once that is addressed then this is good to go! 😀

client/src/views/Navbar.vue
vperuzzi: Yes, you're right! I forgot there really should be a "Components" folder. Fixing that :)

aneuhold (approved):
Looks great! 🏄‍♂️

aschnapp (approved):

PR #7, opened 2020-10-19
Us50 t86 add plot test

aneuhold:
This PR adds a test to the `/plot` endpoint and makes sure an image is successfully encoded and decoded after passing through the API. It also has a lot of code from Michael and Trent that hasn't been merged yet. So this will merge that code unless they put in a PR first. 

This PR also completes [Task 86](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-86).


aschnapp (approved):

aneuhold (commented):
After looking through the changes that I did not submit, it looks good as well!

PR #11, opened 2020-10-24
US 90 - Add CosmologyForm

aneuhold:
## Overview

This PR implements a form, the Cosmology form, into the front-end. It completes US 90. The form looks like so at the moment:

Form interface

It has a drop down for selecting the type of cosmology, which automatically updates the values in each field for the defaults for that type. 

The defaults are brought in from the server, where a new endpoint was created called `/constants`. 

Also a re-usable component was added on the front-end that is called `FormNumberField.vue` which can be used to setup your number fields with error handling in an easy way. 

### App Model Data

An initial idea for the app model data was added to the `Home.vue` component. 

This had a narrow view of only the cosmo constants, so feel free to modify as needed. But other code might need to be modified if the hierarchy level of `cosmo` is changed because the code depends on where it is at the moment. 

## Includes:
- client/package-lock.json 
- client/package.json 
- client/src/App.vue 
- client/src/components/CosmologyForm.vue 
- client/src/components/FormNumberField.vue 
- client/src/router/index.js 
- client/src/views/Home.vue 
- server/requirements.txt 

## Developer Checklist:
- [x] My code compiles and runs locally
- [x] The code follows the Quality Policy file

## Notes:
- There is an error that keeps coming up for `md-select` when running the application. See the screenshot here: Screenshot of error There doesn't seem to be a way to fix it at the moment. It was attempted to move the vue-material version back to beta 11 and to beta 14 but the error still came up. So having to leave it for now. This error is referenced in [this issue on the vue-material repo](https://github.com/vuematerial/vue-material/issues/2270).

### Refs:
- [US 90](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-90)

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -14,30 +14,25 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.11.0",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.11.0.tgz",
-      "integrity": "sha512-TPSvJfv73ng0pfnEOh17bYMPQbI95+nGWc71Ss4vZdRBHTDqmM9Z8ZV4rYz8Ks7sfzc95n30k6ODIq5UGnXcYQ==",
-      "dev": true,
-      "requires": {
-        "browserslist": "^4.12.0",
-        "invariant": "^2.2.4",
-        "semver": "^5.5.0"
-      }
+      "version": "7.12.1",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.12.1.tgz",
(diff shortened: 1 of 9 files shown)


vperuzzi (approved):

Tengelma (approved):
This looks good! Worth noting for eventual refactor that we agreed to axios for network requests & doing mounted instead of created is a better place for a lot of these lifecycle hooks but this is functional so I'm happy to send it through.

aneuhold:
@Tengelma you are absolutely right. I even added the package in this PR but couldn't figure out how to change the header for CORS quick enough so just went to fetch lol. That will be changed.

PR #13, opened 2020-10-30
Task139 halo mod constants

Tengelma:
This is a basic update which provides backend constants to the front end. 

Import within component script via:
import BACKEND_CONSTANTS from 'src/constants/backend_constants';

and use with:
BACKEND_CONSTANTS.<Your Desired Sub-Object>


vperuzzi (approved):

aneuhold (approved):
Looks good! Much more logical layout of the params. This will be great to use from now on for our data modeling needs in the forms.

PR #17, opened 2020-11-02
US92 Halo Model Form

dblevin1:
### Overview

This PR implements the Halo Model form, and satisfies US 92. I added functionality to FormNumberField so it can handle all input types but I decided to make it into a new file to avoid conflicts. This new file is InputField.

### Includes:
- client/src/views/Home.vue 

### Developer Checklist:


client/src/components/HaloModelForm.vue
```diff
@@ -0,0 +1,135 @@
...
+            :currentValue="input.value"
+            :min="input.min"
+            :max="input.max"
+            :inputName="inputName"
+            :inputType="input.inputType"
+            :options="input.options"
+            :setCurrentValue="createSetCurrentValueFunc(inputName)"
+            />
+            <!--<span class="md-error">Please input a valid number</span>-->
+          <!--</md-field>-->
```
aneuhold, line 25: Seems like extra comments. Not a big deal, but can't tell if this is needed anymore.

client/src/components/HaloModelForm.vue
```diff
@@ -0,0 +1,135 @@
...
+        this.setForm(this.formValues);
+      };
+    },
+    // works for scientific notation, negative number, and decimals
+    // Valid: 1, 1.0, -2.0e-8, etc
+    // Invalid: -1e1.0, -e00001, etc
+    isNotNum(num) {
+      return (
+        num.replace(/-?(?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+-]?\d+)?/g, '') !== ''
+      );
```
aneuhold, line 115: This doesn't seem to be used anywhere at the moment? Maybe it is for the future. This seems to be achieved with number parsing from the `Number` class built into JavaScript as well though.

client/src/components/InputField.vue
```diff
@@ -0,0 +1,140 @@
...
+        this.inputStr = this.currentValue;
+      }
+    },
+  },
+  methods: {
+    handleInput(value) {
+      if (typeof value === 'string' || value instanceof String) {
+        this.inputIsInvalid = false;
+        this.setCurrentValue(value);
+      } else if (!Number.isNaN(value)) {
```
aneuhold, line 94: At the moment when values are entered into the `Input` component they don't seem to show an error when they aren't correct. See below for a screenshot:

Screenshot

This might be because of this section of code. Right now the type of the value coming from the inputs will always be a string, so it will never error.

aneuhold (changes requested):
The new `InputField` component looks awesome! Just a change or two and I would be happy to start integrating it into the Forms i had built before. 

Also very beautiful layout of the data for your form. Looks great!

client/src/components/HaloModelForm.vue
dblevin1: Yes this is currently not used anywhere in this form but that regex pattern also matches scientific notation. So I may add it to InputField for better number validation. I currently used that function in some of my other forms it seems to work quite well. I'll remove it from this form though.

client/src/components/InputField.vue
dblevin1: Good to know I didn't realize they all returned strings.

dblevin1: image
fixed

aschnapp (approved):

PR #18, opened 2020-11-02
HM 148 integrate forms into /create

aschnapp:
### Overview

This PR integrates existing forms into /create
I moved the HTTP requests into create because there is no need for it to be in the home component. 
In the next sprint, we'll need to discuss where and how the data is going to be stored. It does not make sense for the server to be called in the /Home route because that is not where it's going to be used. 

As for where to store the data, I believe indexedDB is the best place. Create a new one in /create, then store it in the browser cache. Navigate to home and grab it from cache again to render. If you want to update, call the route /update/<model_id> and /update component grabs it from the cache and updates it. 

To test this PR, please navigate to localhost:8080/create

This pr will also create the UI for [HM-98](https://the-halo-mod.atlassian.net/browse/HM-98)

### Includes:
Please see files changed for this. A lot has changed

### Developer Checklist:


dblevin1 (approved):
Looks and runs good!

dblevin1 (approved):
New changes look good

PR #20, opened 2020-11-07
US90 Cosmology Form v2

aneuhold:
### Overview
This PR updates `CosmologyForm` to make use of `v-model` concepts from it's parent. This way the parent can use `CosmologyForm` with `v-model` and everything will work smoothly. 

This PR also makes some changes to `InputField` so that it makes use of the `v-model` concepts. Although CosmologyForm was adjusted to use `DoubleField`.

`FormNumberField` was deleted because `InputField` replaces it's functionality now. Any component that was using it had that component swapped out for `InputField`. The conversion only involved changing one passed attribute. 

`DoubleField` was changed in two relatively small ways, that do not impact existing usage at all by other components. Even though that is the case, this can be reverted and some other solution could potentially be found if that is not okay. 
- `htmlParam` was added for this situation where html notation is used to indicate a variable. As done on the normal HaloMod site.
- `value` is watched so when it changes without the user directly typing in something into the input, it still updates the input. Like the case where a model is changed.

### Includes:
- client/package-lock.json 
- client/src/components/CosmologyForm.vue 
- client/src/components/FilterForm.vue 
- client/src/components/FormNumberField.vue 
- client/src/components/HaloProfileForm.vue 
- client/src/components/TransferForm.vue 
- client/src/views/Create.vue 
- client/.eslintrc.js 

### Developer Checklist:
- N/A My code has been developer-tested and includes unit tests

### Notes:
- Rule exception added to `.eslintrc` for camelcase variables. This can be taken out if wanted, possibly a workaround could be found if that is the case.

### Refs:
- [US 90](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-90)


vperuzzi (approved):

client/src/components/CosmologyForm.vue
```diff
@@ -13,90 +13,138 @@
...
+        }
+
+        // Set the value in allCosmoData
+        this.allCosmoData[this.cosmoData.cosmo_model][inputType] = newValue;
 
         // Set the value for good.
-        this.setCosmo(this.cosmoValues);
+        this.$emit('updateCosmo', this.cosmoData);
       };
     },
```
Tengelma, line 132: So the way this is set up currently, you're passing the cosmo object into your component by reference then setting the state of its values manually. v-model never actually does anything because by the time you hit your $emit call you've already set those data members in the parent.

One good solution to this is maintaining a local copy of params and sending a copy of that via the emit instead. This made me realize that my forms all need a minor tweak to avoid passing the actual local object up by reference, too.

client/src/constants/backend_constants.js
```diff
@@ -36,6 +36,48 @@ const BACKEND_CONSTANTS = {
...
   tracer_density: null,
+  cosmo_params: {
+    Planck13: {
+      H0: 67.77,
+      Ob0: 0.048252,
+      Om0: 0.30712,
+      z: 0.0,
+      n: 0.9619,
+      sigma_8: 0.8347,
+    },
```
Tengelma, line 47: Nice to see these! One thing you could do that might help simplify a lot of your initialization & changed option logic is keeping these structured like you expect them to be in the payload.

If you have:
{Planck13: {
      z: 0.0,
      n: 0.9619,
      sigma_8: 0.8347,
      cosmo_params: {
        H0: 67.77,
        Ob0: 0.048252,
        Om0: 0.30712,
      }, ...
}

a lot of the manual work you do currently is just a spread operator away!

Not a big deal if you don't decide to change -- just a suggestion to make it a little cleaner.

client/src/components/CosmologyForm.vue
```diff
@@ -13,90 +13,138 @@
...
       v-for="(input, inputName) in inputs"
-      :labelHtml="input.html"
+      :htmlParam="input.html"
       :key="inputName"
       :step="input.step"
-      :currentValue="cosmoValues[inputName]"
+      :value="cosmoData.cosmo_params[inputName] !== undefined
+        ? cosmoData.cosmo_params[inputName]
+        : cosmoData[inputName]"
+      v-on:input="createSetCurrentValueFunc(inputName)($event)"
```
Tengelma, line 24: The logic you use for :value could be used for v-model and you wouldn't need your createSetCurrentValueFunc function at all because it would connect everything without any trouble -- you just might need to watch the object that contains it and update your allCosmoData to support that.

Tengelma (changes requested):

client/src/components/CosmologyForm.vue
aneuhold: Setting a prop manually sets it in the parent? This can be changed to what is being suggested, but that doesn't seem right. https://stackoverflow.com/questions/43701175/is-it-possible-to-change-props-value-from-method-in-vue-component#:~:text=1%20Answer&text=What%20you%20are%20doing%20will,Vue%20(in%20the%20console).&text=The%20value%20will%20actually%20change,re%2Drenders%20for%20any%20reason.

Tengelma: If it's a value it will pass a copy. If its an object / array it will pass by reference. Functions parameters work the same way in javascript!

https://vuejs.org/v2/guide/components-props.html
image

client/src/constants/backend_constants.js
aneuhold: Updated to match this.

client/src/components/CosmologyForm.vue
aneuhold: How might that be implemented in less code? Having a hard time visualizing that.

Tengelma: You would have

v-model="cosmoData.cosmo_params[inputName] !== undefined
        ? cosmoData.cosmo_params[inputName]
        : cosmoData[inputName]"

instead of :value & v-on:input. This connects the DoubleField to whatever field in cosmoData needs it. 

then down below you could have 

watch: {
  cosmoData: {
    deep=True,
    handler() {
       this.allCosmoData[this.cosmoData.cosmo_model] = this.cosmoData;
       this.$emit('updateCosmo', this.cosmoData);
    }, 
  }
}

Tengelma: It's quite possible that v-model won't even accept a computed property like that, though -- I'm not completely sure! I will go ahead and approve but if you feel like tinkering and figure out if it works lemme know haha

Tengelma (approved):
Looks good to me now! Thanks for considering / adopting some of the changes I suggested.

vperuzzi (approved):
I apologize for the delay on this. The form functions identically to the one currently in production. I was unable to identify any errors. Approved.

PR #21, opened 2020-11-08
US93 Transfer Form v2

aneuhold:
### Overview
Includes updates to the `TransferForm` component so that it can accept `v-model` style usage.

### Includes:
- client/src/components/TransferForm.vue 
- client/src/views/Create.vue

### Developer Checklist:
- N/A My code has been developer-tested and includes unit tests

### Refs:
- [US 93](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-93)


vperuzzi (approved):

client/src/components/TransferForm.vue
```diff
@@ -76,20 +98,27 @@ export default {
...
+        this.transferData[varName] = newValue;
 
         // Set the value for good.
-        this.setTransferParams(this.transferParams);
+        this.$emit('updateTransfer', this.transferData);
       };
     },
+    doSomething() {
+      console.log('rendered');
+    },
```
Tengelma, line 110: Probably not intended for this to be here?

client/src/components/TransferForm.vue
```diff
@@ -1,5 +1,7 @@
 <template>
   <div>
+    <p>allTransferData is {{allTransferData}}</p>
+    <p>transferData is {{transferData}}</p>
```
Tengelma, line 4: Survived from your debugging

client/src/components/TransferForm.vue
```diff
@@ -17,18 +19,32 @@
...
+      v-model="transferData.lnk_min"
+    />
+    <DoubleField
+      :param="'lnk_max'"
+      v-model="transferData.lnk_max"
+    />
+    <DoubleField
+      :param="'dlnk'"
+      v-model="transferData.dlnk"
+    />
```
Tengelma, line 45: These all have mins/maxes defined in forms.py from original halomod app. Would be good to have that validation.

client/src/components/TransferForm.vue
```diff
@@ -76,20 +98,27 @@ export default {
     createSetCurrentValueFunc(transferType, varName) {
       return (newValue) => {
         // Set the new value temporarily
-        this.transferParams[transferType][varName] = newValue;
+        this.allTransferData[transferType][varName] = newValue;
+        this.transferData[varName] = newValue;
```
Tengelma, line 102: Looks like you're modifying parent's state again from the child -- same feedback as Cosmo form.

Tengelma (changes requested):
Handful of issues. All easy fixes, though. Will approve after they're fixed.

client/src/components/TransferForm.vue
aneuhold: Good find

client/src/components/TransferForm.vue
aneuhold: Sure did haha

client/src/components/TransferForm.vue
aneuhold: Validation was added.

client/src/components/TransferForm.vue
aneuhold: What you're saying makes sense. Should it be copied into a local object then? Calling emit seems like a good idea because of how Vue turns each piece of state into a relatively complex observable. Copying into another local object seemed like extra memory usage / complexity when it could temporarily be set where it is at and then triggered properly with `$emit`. It can be changed to whatever is preferred though.

Tengelma: Important to think about scale for a second -- even if we maintained 10 copies of the entire payload we would still be working with less than 150kb of memory. The memory concerns aren't really significant because our application state is so shallow / small.

It may seem like we're adding an extra layer of complexity but we do get something out of it! We have some assurance that we're not modifying our parent or child's state until the exact instant we intend to (the emit call or a watch on the prop) and that the components can easily be decoupled if we need to. It's kind of just a separation of concerns -- parent and child can do whatever they want internally without breaking the other. 

As it is the emit call doesn't do anything because it passes a reference to the parent's object up to the parent and asks it to save that reference in the same place. You could remove it and everything related to v-model but keep the prop and everything would be functionally identical -- just not particularly great state encapsulation.

Tengelma (approved):
Looks good! Thanks for making the changes!

PR #22, opened 2020-11-08
Task156 Major code cleanup and final data structure determination

Tengelma:
**Overview**

This includes major cleanup for Home.vue & Create.vue views. It also defines the expected structure for your data and establishes how v-model should work with the Halo Exclusion form as an example. Once this is pulled, everyone should adapt their code to fit this pattern so we can remain consistent.

Feel free to ask questions if anything is unclear.

**Developer Checklist:**

- [X]  The code follows the Quality Policy file on Google Drive


aschnapp (approved):
Looks good 👍

vperuzzi (approved):

client/src/constants/initial_state.json
```diff
@@ -0,0 +1,59 @@
...
+  "tracer_profile": {
+    "tracer_profile_model": null
+  },
+  "halo_profile": {
+    "halo_profile_model": null
+  },
+  "exclusion": {
+    "exclusion_model": null
+  }
+}
```
aneuhold, line 59: If doing it this way, we are missing an opportunity to pre-fill values with constants when the component is built. This pull request doesn't seem to fix that issue, so it breaks existing functionality.

client/src/views/Create.vue
```diff
@@ -28,137 +30,36 @@
...
-import TransferForm from '@/components/TransferForm.vue';
-import FilterForm from '@/components/FilterForm.vue';
 import HaloExclusion from '@/components/HaloExclusion.vue';
-import HaloProfileForm from '@/components/HaloProfileForm.vue';
-import HaloModelForm from '@/components/HaloModelForm.vue';
-import constants from '@/constants/backend_constants';
-import baseURL from '@/env';
-
-const debug = Debug('Create.vue');
+import Blank from '@/constants/initial_state.json';
```
aneuhold, line 35: `Blank` seems like a vague variable name. Can we use something like `initialState`? Plus it is an object and not a class so the capital letter seems odd.

client/src/router/index.js
```diff
@@ -13,6 +13,8 @@ import {
   MdList,
   MdApp,
   MdCheckbox,
+  MdSubheader,
```
aneuhold, line 16: This fixes an error that has been around for a while 😄 Awesome 👍

aneuhold (changes requested):
This PR removes functionality and doesn't replace it or try to replace it. It deletes a lot of work and puts in one example of something that everyone should change their code to. This means that anything people are working on needs to be changed now. This is more of an order to the team than a PR to be honest. If the existing code was changed to work with this new model, then it would make sense. Otherwise it is a large task for everyone to go back and change what they were doing once again.

client/src/views/Create.vue
Tengelma: Fixed in recent commit

aneuhold (approved):

PR #25, opened 2020-11-12
Hm104 tracer concentration

aschnapp:
Overview
This implements Tracer Concentration form. It will be refactored when I work on halo concentration to support both. This also implements lodash.clonedeep instead of using the really expensive JSON serialization cloning technique. 

Includes:
Please check files changed for changes

Developer Checklist:


client/src/components/TracerConcentration.vue
```diff
@@ -0,0 +1,124 @@
...
+        case ('Ludlow16'):
+          return form === 'f' || form === 'C';
+        case ('Ludlow16Empirical'):
+          return form === 'c0_0' || form === 'c0_z'
+            || form === 'beta_0'
+            || form === 'beta_z'
+            || form === 'gamma1_0'
+            || form === 'gamma1_z'
+            || form === 'gamma2_0'
+            || form === 'gamma2_z';
```
Tengelma, line 94: I don't think these options should be hardcoded. These are all available in backendconstants via the CMRelation_params and there's no need to do this manually.

``

client/src/components/TracerConcentration.vue
```diff
@@ -0,0 +1,124 @@
...
+    emit() {
+      this.$emit('onChange', {
+        tracer_concentration_model: this.model.concentration_select,
+        tracer_concentration_params: {
+          ...this.additional_values.fields,
+          sample: this.additional_values.selection.select,
+        },
+      });
+    },
+  },
```
Tengelma, line 108: If you make the 'model' object in your data consistent with what the parent component expects and just write to tracer_concentration_model & tracer_concentration_params directly (see BiasForm & HODForm) you can get rid of this function completely.

client/src/components/TracerConcentration.vue
```diff
@@ -0,0 +1,124 @@
...
+      handler() {
+        this.emit();
+      },
+    },
+    additional_values: {
+      deep: true,
+      handler() {
+        this.emit();
+      },
+    },
```
Tengelma, line 121: Same thing here. You don't need two watches if you just watch the model object which contains everything you need.

client/src/components/TracerConcentration.vue
Tengelma, line 108: If you make the 'model' object in your data consistent with what the parent component expects and just write to tracer_concentration_model & tracer_concentration_params directly (see BiasForm & HODForm) you can get rid of this function completely.

client/src/components/TracerConcentration.vue
```diff
@@ -0,0 +1,124 @@
...
+          return form === 'f' || form === 'C';
+        case ('Ludlow16Empirical'):
+          return form === 'c0_0' || form === 'c0_z'
+            || form === 'beta_0'
+            || form === 'beta_z'
+            || form === 'gamma1_0'
+            || form === 'gamma1_z'
+            || form === 'gamma2_0'
+            || form === 'gamma2_z';
+        default:
```
Tengelma, line 95: I don't think these options should be hardcoded. These are all available in backendconstants via the CMRelation_params and there's no need to do this manually when you can draw from that (see BiasForm & HODForm)

client/src/constants/form_choices.json
```diff
@@ -0,0 +1,43 @@
+{
+  "concentration": {
+    "main": {
+      "Bullock01": "Bullock (2001) Physical Form",
+      "Bullock01Power": "Bullock (2001) Power-Law",
+      "Duffy08": "Duffy (2008) Power-Law",
+      "Zehavi11": "Zehavi (2011) Power-Law",
+      "Ludlow16": "Ludlow (2016)",
+      "Ludlow16Empirical": "Ludlow (2016) Empirical"
+    },
```
Tengelma, line 10: I'm fine with extracting these but we should keep our constants in one file. Can these be moved to backend constants?

client/src/constants/form_choices.json
```diff
@@ -0,0 +1,43 @@
...
+      "c0_0": 3.395,
+      "c0_z": -0.215,
+      "beta_0": 0.307,
+      "beta_z": 0.54,
+      "gamma1_0": 0.628,
+      "gamma1_z": -0.047,
+      "gamma2_0": 0.317,
+      "gamma2_z": -0.893
+    }
+  },
```
Tengelma, line 33: These are just copies of values in backendconstants! Draw from that and they actually have the structure you need, already, instead of you having to marshal things in manually.

client/src/components/TracerConcentration.vue
```diff
@@ -0,0 +1,124 @@
...
+        />
+      </div>
+      <md-field v-if="additionalFormControl('sample')">
+        <label>sample</label>
+        <md-select v-model="additional_values.selection.select">
+          <div v-for="(value, key) in additional_values.selection.values" :key="key">
+            <md-option :value="key">{{value}}</md-option>
+          </div>
+        </md-select>
+      </md-field>
```
Tengelma, line 32: It would be great if this field was tied more explicitly to the model where the exception occurs. Its not really clear what additionalFormControl does or is expected to return here unless I go and look at the function.

Tengelma (changes requested):
I think you made this way harder than it needed to be for yourself! You can simplify this code dramatically by using data models we already have access to and I would suggest you do so!

Tengelma (approved):
Addressed everything I asked you to address. And clever solution to the "sample" problem!

Approved.

vperuzzi (approved):

PR #26, opened 2020-11-13
US-95-TracerProfileForm

vperuzzi:
Overview
This implements Tracer Profile form. See form screenshot from original HaloMod:
image

Includes:
Please check files changed for changes

**Developer Checklist:**

image


client/src/components/TracerProfileForm.vue
```diff
@@ -0,0 +1,86 @@
...
+        :value="value"
+        :param="param"
+        range=false
+        :placeholder="String(defaults[param])"
+        v-model="model.profile_params[param]"/>
+      <md-checkbox
+        v-if="typeof value === 'boolean'"
+        class="md-primary"
+        v-model="model.profile_params[param]"
+      >
```
Tengelma, line 29: This has a little bug ( try typing f into the text field ) mainly because doublefield saves values as a string for validation purposes. 

Simple fix would be to do v-if="param === 'interop'" for the checkbox and v-else otherwise.

client/src/views/Create.vue
```diff
@@ -70,6 +72,10 @@ export default {
           component: HODForm,
           model: 'hod',
         },
+        {
+          component: TracerProfileForm,
+          model: 'tracerProfile',
+        },
```
Tengelma, line 78: Per initial_state.json this should have tracer_profile as model. Could run into issues with overwrites during unpacking if naming convention isn't exact.

Tengelma (changes requested):
Looking forward to when you copy paste this and Ctrl+f replace all for HaloProfile lol

client/src/views/Create.vue
vperuzzi: Yep, I missed that. Thank you.

aschnapp (changes requested):
After you fix conflicts, I'll approve. 

You're missing the default values though.

aneuhold (approved):

Tengelma (approved):
Looks like you fixed the issues pointed out previously. Approved!

aschnapp (approved):

PR #35, opened 2020-11-21
US92 halo model form

dblevin1:
### Overview
Quick fix so the HaloModel form returns the correct model with only the parameter values.

### Includes:

### Developer Checklist:

### DIFF
--- a/client/src/components/HaloModelForm.vue
+++ b/client/src/components/HaloModelForm.vue
@@ -4,7 +4,7 @@
       <div class="md-layout-item">
         <md-field>
           <label>Halo Centre Spectrum</label>
-          <md-select v-model="model.halo_model" md-dense>
+          <md-select v-model="model.hc_spectrum" md-dense>
             <md-option
               v-for="(value, choice) in choices"
               :key="choice"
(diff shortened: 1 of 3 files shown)


aschnapp (approved):

aschnapp (changes requested):
It's still failing

Tengelma:
> It's still failing

The issue is with the 'range' fields -- these need to have a min and a max and those need to correspond to something in the payload. I just made a commit on your US92 branch that fixes it I think but it uses two separate fields instead of 1 for both of the ranges & also doesn't account for the log scale.

Just as an example, the field from forms.py I listed below corresponds to 'rmin' and 'rmax' in the payload so if you have a range slider it needs to have two ends. The reverse of that 'np.log10' call also needs to be done on it before its actually sent in the payload. We can probably send this through to get dev back in working order but this functionality should probably be handled at some point :)

image

aschnapp (approved):
Working now!

aneuhold (approved):
Looks good!

aschnapp:
@dblevin1 can you look at this pr before you merge? I moved the data to backend constant.
https://github.com/steven-murray/TheHaloMod-SPA/pull/36

PR #36, opened 2020-11-22
Us92 move data to backend constants

aschnapp:


aneuhold:
`backend_constants` was generated by Steven's script. So it might not be good to assume that it will have a different form than what it has now in case it is implemented later as an automatically retrieved set of values. 

@Tengelma comment?

Tengelma:
> `backend_constants` was generated by Steven's script. So it might not be good to assume that it will have a different form than what it has now in case it is implemented later as an automatically retrieved set of values.
> 
> @Tengelma comment?

I agree with @aneuhold. Part of the reason its named what it is is because everything in there can be pulled directly from halomod and hmf packages. We're eventually trying to draw these values directly from server as part of our setup so form specific details like input types / ranges / etc... shouldn't really be here.

@aschnapp this is something we talked about when trying to figure out a home for the model_options dict -- I'm actually inclined to think we should stick that (and other stuff related to field initialization) somewhere else.

PR #37, opened 2020-11-22
HM169 front end data layer, second iteration, future refactor expected

aschnapp:
Overview
Uses indexed db for front end data layer. 
It grabs data out of the db on at /home, and adds or updates it into the the db in /create. 
@Tengelma please add a check for model name in model metadata form to check if the model name exists with the keys() function.

This is partially done, you can start using this for /update/{plot_name}. DO NOT USE THE PLOT!

Includes: 
Check Files Changed

Developer Checklist:


Tengelma:
Could you provide a little documentation on usage? Specifically, what should I be calling (and with what params) to
- create a model
- update a model
- delete a model
- get a list of available models
from the browser?

I'm sure I can figure it out with a bit of time but should probably be included to save everyone a bit of time.

aschnapp:
@Tengelma I added some comments into IDB.js, but it's quite clear at the moment already.

Tengelma (commented):
Approving because it works but language used for idb calls should be more expressive. Not immediately clear that keys == model labels & values == param objects. That's mostly what I was getting at with previous comment.

If this is being used exclusively for model / form data we can afford to abstract more and use language that corresponds to the use case instead of shallowly wrapping idb functionality.

aschnapp:
@Tengelma the abstraction is going to be in another component. It's a waste of indexed db to only use it for model data. 
I'm pushing another component with better documentation. Please check

client/src/main.js
```diff
@@ -3,12 +3,11 @@ import axios from 'axios';
 import App from './App.vue';
 import router from './router';
 import IDB from './utils/IDB';
+import API from './utils/API';
 
 Vue.config.productionTip = false;
 axios.defaults.withCredentials = true;
```
Tengelma, line 9: This line needs to be in API if we're moving the import there.

Tengelma: Referring to the axios.defaults.withCredentials in case that wasn't clear

Tengelma:
And ya I already approved. I just think we just need to be careful not to overuse idb for state management. It makes sense to me in the specific context of restoring a session after closing/crashing a browser but I still think it would be good to hook into a store that's just a javascript object for most of the operations we need. Idk what everyone else thinks though.

Tengelma (approved):

aschnapp:
I only forsee 3 places where IDB is going to be called at this moment, once in home and twice in create (mounted and before redirect back to /home). We might use it for other things down the line like user preferences though.

client/src/components/SubmitButton.vue
```diff
@@ -24,7 +24,7 @@
 </template>
 
 <script>
-import baseurl from '../env';
+// import baseurl from '@/env';
```
aneuhold, line 27: This can be removed.

client/src/utils/API.js
```diff
@@ -0,0 +1,52 @@
...
+      });
+
+      // puts model inside idb after successful api call
+      await this.db.put(label, {
+        name: label,
+        model,
+      });
+    } catch (error) {
+      console.error(error);
+    }
```
aneuhold, line 50: This looks great. The documentation with types is greatly appreciated 😄

client/src/views/Create.vue
```diff
@@ -142,10 +142,10 @@ export default {
...
           model: 'cosmo',
         },
-        {
-          component: HaloModelForm,
-          model: 'halo_model',
-        },
+        // {
+        //   component: HaloModelForm,
+        //   model: 'halo_model',
+        // },
```
aneuhold, line 148: Why was this one removed?

client/src/views/Home.vue
```diff
@@ -10,6 +10,13 @@ export default {
   name: 'Home',
   // TODO: Make the data at the app level and learn how to pass this to routes
   data: () => ({}),
+  async created() {
+    try {
+      console.log(await this.$db.delete('Model'));
+    } catch (e) {
+      console.log(e);
+    }
+  },
```
aneuhold, line 19: Is this still functional? Doesn't seem that `$db` is declared on the Vue instance anymore.

aneuhold (changes requested):
Some comments, some questions.

client/src/utils/API.js
aschnapp: made it even better :)

client/src/views/Create.vue
aschnapp: It crashes the server right now

client/src/views/Home.vue
aschnapp: Yes, I added this.$db back now. We need it.

client/.eslintrc.js
```diff
@@ -13,6 +13,7 @@ module.exports = {
   rules: {
     'no-console': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
     'no-debugger': process.env.NODE_ENV === 'production' ? 'warn' : 'off',
+    'consistent-return': 'off',
```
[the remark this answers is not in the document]
aschnapp: Turned it back on

client/src/utils/Store.js
```diff
@@ -0,0 +1,126 @@
...
+  cloneModel = async (model, name) => {
+    try {
+      // unknown api call somewhere,
+      // name may be old name for api call
+      // use getModel to create the cloned model
+      await Promise.all([this.setModel(model, name), this.createPlot()]);
+    } catch (error) {
+      console.error(error);
+    }
+  }
```
Tengelma, line 97: Here's the server endpoint

image

client/src/utils/Store.js
```diff
@@ -0,0 +1,126 @@
...
+  updateModel = async (model, name) => {
+    try {
+      /* some api call here */
+      await Promise.all([this.setModel(model, name), this.createPlot()]);
+    } catch (error) {
+      console.error(error);
+      // better error handling here, some vue event?
+    }
+  }
+
```
Tengelma, line 82: Relevant server endpoint:

image

client/src/utils/Store.js
```diff
@@ -0,0 +1,126 @@
...
+   * @param {String} name the name to set the model
+   */
+  setModel = async (model, name) => {
+    // puts model inside idb after successful api call
+    await this.db.put(name, {
+      name,
+      model,
+    });
+    this.models[name] = model;
+    this.keys = this.db.keys();
```
Tengelma, line 124: Should probably grab keys from this.models instead of this.db (faster since its a js object)

Tengelma: In fact now that I'm thinking about it it may better to maintain the specific array @ this.keys and mutating it instead of overriding it with a reference to a new array whenever it needs to be adjusted. That lets us keep a reference to that object in home and prevents us from needing a call to keys whenever we do anything.

adding to an array is just keys.push()
removing from it would require a helper function
```
function removeIitem(arr, value) {
  var index = arr.indexOf(value);
  if (index > -1) {
    arr.splice(index, 1);
  }
  return arr;
}
```

Create/clone would add to it.
Update wouldn't do anything with it.
Delete would remove from it.

Tengelma (commented):
I really like these changes and think they make this data layer much more usable. I attached some screenshots of the relevant server endpoints in places where you haven't implemented it yet (just so you wouldn't have to go looking).

Also some little comments here and there to make it even better.

client/src/views/Create.vue
aneuhold: It doesn't seem to be crashing anything when running it locally with that included. Does doing something specific cause it to crash?

client/src/utils/Store.js
aschnapp: You're doing two for loops to get that, might as well just get it from the db. The keys() in the async call is constant time.

aschnapp: it's already in the models, I save the names inside the models object.  The keys just gets an array.

aneuhold:
Once the API calls that @Tengelma mentioned are implemented, I am happy approving this.

client/src/utils/Store.js
```diff
@@ -0,0 +1,162 @@
...
+        fig_type,
+        img_type: 'png',
+      });
+      this.plot = `data:image/png;base64,${data.figure}`;
+      return this.plot;
+    } catch (error) {
+      console.error(error);
+      return null;
+    }
+  }
```
Tengelma, line 61: Would be good if this was called any time state updates (create/update/clone/delete) instead of having to lazily request it. Plot choice isn't specific to each model so this could have a home, here, in state as a global context.

Tengelma (commented):
Had a couple comments but overall this is looking great. Looking forward to see it integrated w/ UI.

aneuhold (approved):
Looks great!

client/src/utils/Store.js
```diff
@@ -0,0 +1,162 @@
...
+   * @param {Object} model model to flatten
+   * @returns {Object} flattened params
+   */
+  flatten = (model) => {
+    const params = {};
+    Object.values(model).forEach((value) => {
+      Object.assign(params, value);
+    });
+    return params;
+  }
```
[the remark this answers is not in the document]
aschnapp: No. I changed it to assign because you're creating a new object every time you using the spread. When you use assign, you're writing it into the same params object. 
params = { ...params, ...value } vs Object.assign(params, value)

client/src/utils/Store.js
```diff
@@ -0,0 +1,162 @@
...
+   * initializes cache
+   */
+  init = async () => {
+    const k = await keys();
+    this.models = new Map();
+    k.forEach((key) => {
+      this.models.set(key, get(key));
+    });
+    await Promise.all(this.models);
+  }
```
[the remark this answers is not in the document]
aschnapp: Yeah we can do that. It'll sync it everytime it initializes.

client/src/utils/Store.js
aschnapp: createPlot is called in create, update, and clone. I dont know what you mean here?

client/src/utils/Store.js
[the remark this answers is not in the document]
Tengelma: Imo this is less expressive for no payoff but ok

client/src/utils/Store.js
Tengelma: If so then we don't need this right?

` // should redirect to home. Image should be requested when at home component \
this.image = await this.$store.createPlot(this.meta.fig_type);`

Tengelma (approved):
This looks good!

aschnapp:
@Tengelma I'm going to merge it for now. As written in slack, whoever is going to be using it can decide where createPlot is going to be called. I'm going to leave that open for future changes.

PR #38, opened 2020-11-29
Minor changes to support store usability by front end

Tengelma:
Hi this is a followup to Michael's pull request that ever so slightly restructures how data is stored (using an object instead of a map)

These changes let any vue component save a reference to the stores state locally like
`data: { READ_ONLY: this.$store.state }`
and then use values within for v-binds & computed components.

As an example, if I want to v-bind the src of an image tag to the thing stored in the store, I just have 
``

or if I want to always maintain current keys in v-for I can:
```
<div v-for="name in READ_ONLY.models.keys()">
  Do stuff.
<div>
```

Be mindful that you should never modify this state object directly -- consider it read only because the store does a lot of work to keep browser, server and cache in sync. Use functions from the store class any time you need to make a mutation.

### DIFF
--- a/client/src/utils/Store.js
+++ b/client/src/utils/Store.js
@@ -11,8 +11,10 @@ import {
 axios.defaults.withCredentials = true;
 export default class API {
   constructor() {
-    this.plot = '';
-    this.models = new Map();
+    this.state = {
+      plot: '',
+      models: {},
+    };
   }
 
   /**
@@ -20,10 +22,13 @@ export default class API {
    */
   init = async () => {
     const k = await keys();
+    const models = new Map();
     k.forEach((key) => {
-      this.models.set(key, get(key));
+      const obj = get(key);
+      models.set(key, obj);
+      this.state.models[key] = obj;
     });
-    await Promise.all(this.models);
+    await Promise.all(models);
   }
 
   /**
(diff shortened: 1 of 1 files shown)


client/src/utils/Store.js
```diff
@@ -11,19 +11,24 @@ import {
...
    * initializes cache
    */
   init = async () => {
     const k = await keys();
+    const models = new Map();
     k.forEach((key) => {
-      this.models.set(key, get(key));
+      const obj = get(key);
+      models.set(key, obj);
+      this.state.models[key] = obj;
```
aschnapp, line 29: Does this work? Object is a promise. Dont you have to convert models after promise.all?

client/src/utils/Store.js
```diff
@@ -127,7 +130,7 @@ export default class API {
    * @param {String} name the name of the model
    * @returns {Object} A copy of the target model, or null
    */
-  getModel = async (name) => clonedeep(await this?.models?.get(name));
+  getModel = async (name) => clonedeep(await this?.state.models[name]);
```
aschnapp, line 133: state, models are going to be there, no need for the ? anymore

aschnapp (changes requested):

client/src/utils/Store.js
aschnapp: Dont have to change this, just a suggestion.

client/src/utils/Store.js
aschnapp: This isnt working, you need to convert the map into an object after the await Promise.all(). 
Screen Shot 2020-11-29 at 3 13 05 PM

aschnapp (approved):

aneuhold:
It looks like the build is failing at the moment because of an error in not defining `map`.

aneuhold (approved):

PR #41, opened 2020-11-30
US168 Update to state management

Tengelma:
@aneuhold 

I tied the Graph component's image tag to the store's state and deleted all the createPlot calls from index.js -- no need to manage that anymore!

Getting modelNames to be cleaner was a little harder -- ended up creating an array of modelNames in the store's state & updated it in the few places where it needed to be updated in there. Tied your v-for to that in index.js so that should all be good to.

### DIFF
--- a/client/src/components/Graph.vue
+++ b/client/src/components/Graph.vue
@@ -6,7 +6,7 @@
           <h3 class="md-title">Plot</h3>
         </div>
       </div>
-      <img v-if="image !== null" :src="image"/>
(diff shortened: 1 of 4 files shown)


aneuhold:
@Tengelma this looks great! Just fixed a bug with it trying to use an unknown function (left over unneeded plot update call), and it wasn't updating the plot when a model was deleted so just fixed that and it still works fast too. Rock on!

PR #53, opened 2021-01-24
US189 Store Testing

aneuhold:
## Overview
Implements user story 189, Store tests. 

### Addition of Mock Server
A mock server was setup to intercept network requests. This can be used by anyone and makes use of [Mirage JS](https://miragejs.com/). It is pretty easy to learn, and there is an example of how to use it in `Store.spec.js`. But overall it is as easy as doing the following in your tests:

```js
import makeServer from '../mockServer';

describe('Some tests', () => {
  let server;
  beforeAll(() => {
    server = makeServer('test');
  });

  afterAll(() => {
    server.shutdown();
  });

  // ... your tests here
});
```

After doing that, all of your tests will have the network connections intercepted and returned with what you like according to what is in `mockServer.js`.

The mock server does have a super simple DB you can make use of, but no need to if you don't want to. That is what the `models` section is.

### `fake-indexeddb`

A fake `indexeddb` was used for the tests, because during tests there isn't a `window` object which is normally used for `indexeddb` to store things in. Aka, this is a fake `idb-keyval`.

### Code Coverage

Code coverage wasn't getting the `.js` files. It seems to be because there was a space in the definition in `jest.config.js` where it says `<rootDir>/src/**/*.{vue, js}`. Changing it to `<rootDir>/src/**/*.{js,vue}` fixed that. And now we have pretty nice code coverage kinda! 😅

Code coverage

## Includes:
- client/jest.config.js 
- client/package-lock.json 
- client/package.json 
- client/src/utils/Store.js 
- client/tests/example_data/figureData.json 
- client/tests/mockServer.js 

## Developer Checklist:

## Refs:
- [US 189](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-189)


aschnapp (approved):

Tengelma (approved):
Happy with this. Tests existing functionality well but we should probably improve the store generally at some point to have better error handling & test that too. 

Approved!

vperuzzi (approved):

PR #54, opened 2021-01-24
Hm207v2 add create and edit to its own route

aschnapp:
## Overview
moves create and edit into its own routes

## Includes:
please see files changed for changes

## Developer Checklist:
- [] My code has been developer-tested and includes unit tests

## Notes:
- Needs a better progress bar

## Refs:
- [US 207](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-207)


client/src/views/Create.vue
```diff
@@ -176,16 +198,28 @@ export default {
...
-    this.createForms();
+    async handleSave() {
+      this.loading = true;
+      console.log(this.params);
+      if (this.$route.name === 'Edit') {
+        await this.$store.updateModel(this.$route.params.id, this.params);
+        if (this.model_metadata.model_name !== this.$route.params.id) {
+          await this.$store.cloneModel(this.$route.params.id, this.model_metadata.model_name);
+          await this.$store.deleteModel(this.$route.params.id);
+        }
```
Tengelma, line 216: Letting people modify the name of their model from the 'forms' page seems a little weird to me. Maybe, like fig_type, model_name should only be editable from top level (list view of models) and forms present here could just be related to the model itself instead of the metadata?

Tengelma: Also rename functionality should probably be the domain of $store instead of here. Potentially reusable logic.

Tengelma (commented):
I think we have a tendency to just dump all the logic into this one view when the couldbase would be cleaner if we split some things up. It would be great if there were separate views for Edit / Create so we could abstract the route & store related logic into those and keep the forms independent of all of that stuff. Would separate concerns a little bit & make things easier to test down the road.

The basic idea:  
- Make the current Create.vue a FormsComposite.vue and put it in components.
- Create two new files Create.vue & Edit.vue in views and use them as simple wrappers for FormsComposite.vue.
- Pass initial state (either from INITIAL_STATE or $store) to FormsComposite via props & initialize data there from that.  
- Pass model state up from FormsComposite to Create and Edit via v-model (whenever something changes).
- Keep buttons & their corresponding handlers in the wrappers so they can have unique functions for updating vs creating and maybe unique labels, too (create in Create & Save or Update in Edit).

Is that something that makes sense to the rest of you?

client/src/views/Create.vue
aneuhold: @Tengelma ya, rename still needs to be added into `Store`.

aneuhold:
Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:

UI Example

It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.

aschnapp:
> I think we have a tendency to just dump all the logic into this one view when the couldbase would be cleaner if we split some things up. It would be great if there were separate views for Edit / Create so we could abstract the route & store related logic into those and keep the forms independent of all of that stuff. Would separate concerns a little bit & make things easier to test down the road.
> 
> The basic idea:
> 
>     * Make the current Create.vue a FormsComposite.vue and put it in components.
> 
>     * Create two new files Create.vue & Edit.vue in views and use them as simple wrappers for FormsComposite.vue.
> 
>     * Pass initial state (either from INITIAL_STATE or $store) to FormsComposite via props & initialize data there from that.
> 
>     * Pass model state up from FormsComposite to Create and Edit via v-model (whenever something changes).
> 
>     * Keep buttons & their corresponding handlers in the wrappers so they can have unique functions for updating vs creating and maybe unique labels, too (create in Create & Save or Update in Edit).
> 
> 
> Is that something that makes sense to the rest of you?

The differences in create and edit functions can be in done in store, otherwise: prepopulating user created data = prepopulating default data. I dont see why we need an extra wrapper.

client/src/views/Create.vue
aschnapp: The location can be changed later. It's not part of the scope of this change.

aschnapp:
> Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:
> 
> UI Example
> 
> It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.

fixed the bottom styling. I cannot find a better compoent than md-app-toolbar for md-app right now. We can fix the styling later in another location. Maybe change the stuff on the navbar?

Tengelma:
> > Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:
> > UI Example
> > It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.
> 
> fixed the bottom styling. I cannot find a better compoent than md-app-toolbar for md-app right now. We can fix the styling later in another location. Maybe change the stuff on the navbar?

Floating buttons that are semi-transparent normally & solid on hover would work well I think. On the styling front I think we'd also benefit from some more whitespace on the left and right of the forms themselves. Don't think they need to take the entire screen, ya?

aschnapp:
> > > Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:
> > > UI Example
> > > It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.
> > 
> > 
> > fixed the bottom styling. I cannot find a better compoent than md-app-toolbar for md-app right now. We can fix the styling later in another location. Maybe change the stuff on the navbar?
> 
> Floating buttons that are semi-transparent normally & solid on hover would work well I think. On the styling front I think we'd also benefit from some more whitespace on the left and right of the forms themselves. Don't think they need to take the entire screen, ya?

Yeah I agree with all that. We can definately use better styling. Let's do that on a different US.  It can still be in a modal (which supports routing) too if Steven insists on it (which would make sense to keep the toolbar). 

I don't think making it look good is within the scope of this US though. Let's get the routing to work and we can work on a better style in a separate US.

Tengelma:
> > I think we have a tendency to just dump all the logic into this one view when the couldbase would be cleaner if we split some things up. It would be great if there were separate views for Edit / Create so we could abstract the route & store related logic into those and keep the forms independent of all of that stuff. Would separate concerns a little bit & make things easier to test down the road.
> > The basic idea:
> > ```
> > * Make the current Create.vue a FormsComposite.vue and put it in components.
> > 
> > * Create two new files Create.vue & Edit.vue in views and use them as simple wrappers for FormsComposite.vue.
> > 
> > * Pass initial state (either from INITIAL_STATE or $store) to FormsComposite via props & initialize data there from that.
> > 
> > * Pass model state up from FormsComposite to Create and Edit via v-model (whenever something changes).
> > 
> > * Keep buttons & their corresponding handlers in the wrappers so they can have unique functions for updating vs creating and maybe unique labels, too (create in Create & Save or Update in Edit).
> > ```
> > 
> > 
> > Is that something that makes sense to the rest of you?
> 
> The differences in create and edit functions can be in done in store, otherwise: prepopulating user created data = prepopulating default data. I dont see why we need an extra wrapper.

Because it separates form logic from routing / store interaction logic, which makes both ends easier to test & lets us customize the behavior of certain routes more easily. Generally nice to have core side effects separate from presentational stuff. I'm up for making this change if you just don't want to do the legwork haha

Tengelma:
> > > > Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:
> > > > UI Example
> > > > It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.
> > > 
> > > 
> > > fixed the bottom styling. I cannot find a better compoent than md-app-toolbar for md-app right now. We can fix the styling later in another location. Maybe change the stuff on the navbar?
> > 
> > 
> > Floating buttons that are semi-transparent normally & solid on hover would work well I think. On the styling front I think we'd also benefit from some more whitespace on the left and right of the forms themselves. Don't think they need to take the entire screen, ya?
> 
> Yeah I agree with all that. We can definately use better styling. Let's do that on a different US. It can still be in a modal (which supports routing) too if Steven insists on it (which would make sense to keep the toolbar).
> 
> I don't think making it look good is within the scope of this US though. Let's get the routing to work and we can work on a better style in a separate US.

Just concerned that if we present it as is to Steven it will feel like a regression to him. Some route switches are slow and rerender the whole page for me (Home -> Create) and in general it looks worse than the last thing we demoed to him. Would rather spend a little time making everything nice so we can actually sell it to him as an alternative to the 'single page' he was asking for.

I think the style changes you made did something to Home btw. 
image

aschnapp:
> > > > > Right now, in addition to what Trent has mentioned, this is kind of a confusing UI. For example when creating a new model:
> > > > > UI Example
> > > > > It shows the "Save" and "Close" at the top right under the "Home" and "About" menus. Also the bottom of the screen is cut off. If it is meant to feel like a new window that is coming up, then it should likely not include the top bar, rather the top bar should be replaced by a back arrow to the top left and the "Home" and "About" should probably be gone. Also the "Save" and "Close" might be better served at the bottom in this case so as to not confuse the user, because the "Home" and "About" used to be there a second ago.
> > > > 
> > > > 
> > > > fixed the bottom styling. I cannot find a better compoent than md-app-toolbar for md-app right now. We can fix the styling later in another location. Maybe change the stuff on the navbar?
> > > 
> > > 
> > > Floating buttons that are semi-transparent normally & solid on hover would work well I think. On the styling front I think we'd also benefit from some more whitespace on the left and right of the forms themselves. Don't think they need to take the entire screen, ya?
> > 
> > 
> > Yeah I agree with all that. We can definately use better styling. Let's do that on a different US. It can still be in a modal (which supports routing) too if Steven insists on it (which would make sense to keep the toolbar).
> > I don't think making it look good is within the scope of this US though. Let's get the routing to work and we can work on a better style in a separate US.
> 
> Just concerned that if we present it as is to Steven it will feel like a regression to him. Some route switches are slow and rerender the whole page for me (Home -> Create) and in general it looks worse than the last thing we demoed to him. Would rather spend a little time making everything nice so we can actually sell it to him as an alternative to the 'single page' he was asking for.
> 
> I think the style changes you made did something to Home btw.
> image

fixed home. 

as for Steven's concern, we'll get something better afterwards. There are huge changes needed for create/ anyway. All the forms needs to be refactored and modified to support user generated values.

aneuhold:
@aschnapp in the description for this PR, what does "Needs a better progress ba" mean?

aneuhold:
Also, the design seems to be getting further off:

image

The font for the "New Model" part looks to be different than everything else on the page and the spacing is different than the top bar. Also the white space around "Model Metadata" in the left menu is different than every other menu item.

aneuhold:
One more thing: Spacing at the top, under the main menu bar on the home screen looked better before than it does in this branch. Dev spacing currently:

image

The spacing in this branch at the moment:

image

aschnapp:
@Tengelma @aneuhold /home is left as is right now, except navbar is now edited on a per component basis (you can have custom buttons) 

I wanted to keep the nav in its own component, but seems like it isn't possible with md-app... @Tengelma if you have any ideas on how you keep nav inside navbar, lmk. 

all styling issues have been fixed. /home is as it currently is. 

Screen Shot 2021-01-25 at 3 52 16 PM

aneuhold:
@aschnapp the UI looks quite a bit better now.

Just still having a hard time seeing why we need routes in this situation. It seems to be quite slow at the moment, and we still don't have a lot of other functionality included yet, which means it might get slower. Testing routes sounds more complicated than testing components. It could be easier, but it definitely makes the code harder to read and understand. So what are the upsides to routes again? Other people on the team can approve this PR without this question being answered. But until I have a better understanding of the benefits, I don't feel comfortable approving.

aschnapp:
> @aschnapp the UI looks quite a bit better now.
> 
> Just still having a hard time seeing why we need routes in this situation. It seems to be quite slow at the moment, and we still don't have a lot of other functionality included yet, which means it might get slower. Testing routes sounds more complicated than testing components. It could be easier, but it definitely makes the code harder to read and understand. So what are the upsides to routes again? Other people on the team can approve this PR without this question being answered of course. But until I have a better understanding of the benefits, I don't feel comfortable approving.

@aneuhold  try to navigate to localhost:8080/create and see what happens in dev branch

Tengelma:
I'm on board with routes but I don't think this is ready for dev:
- Auto plot updates on Plot Type change is broken.
- Navbars are meant to be navbars. Save/Cancel != Navigation
- Scroll tracking (and menu snapping) isn't working again because intersection line has changed
- Transition to/from Create is super slow when it should be very very fast (like switching between Home & About).
- Transition to/from Create causes entire page to rerender instead of preserving navbar between routes.
- Style still has a lot of room for improvement

Let's just slow down a bit & make it into something that can be stable, functional & look better than the current format. This is going to take some time to get right & there's no reason to rush it through when routes are just dynamic components and this isn't really blocking anything or affecting future development.

aneuhold:
> > @aschnapp the UI looks quite a bit better now.
> > Just still having a hard time seeing why we need routes in this situation. It seems to be quite slow at the moment, and we still don't have a lot of other functionality included yet, which means it might get slower. Testing routes sounds more complicated than testing components. It could be easier, but it definitely makes the code harder to read and understand. So what are the upsides to routes again? Other people on the team can approve this PR without this question being answered of course. But until I have a better understanding of the benefits, I don't feel comfortable approving.
> 
> @aneuhold try to navigate to localhost:8080/create and see what happens in dev branch
> the /create#hash goes nowhere, and if you take "create" out the #hash will still go nowhere. Now when you refresh the page it goes back to where you are.
> I dont notice any difference in speed on my machine though. It's probably just the animation in the dev branch making it feel like it's faster.

The question was more about why we need routes at all, not the difference between this implementation of routes and the one on dev.

Tengelma:
> > > @aschnapp the UI looks quite a bit better now.
> > > Just still having a hard time seeing why we need routes in this situation. It seems to be quite slow at the moment, and we still don't have a lot of other functionality included yet, which means it might get slower. Testing routes sounds more complicated than testing components. It could be easier, but it definitely makes the code harder to read and understand. So what are the upsides to routes again? Other people on the team can approve this PR without this question being answered of course. But until I have a better understanding of the benefits, I don't feel comfortable approving.
> > 
> > 
> > @aneuhold try to navigate to localhost:8080/create and see what happens in dev branch
> > the /create#hash goes nowhere, and if you take "create" out the #hash will still go nowhere. Now when you refresh the page it goes back to where you are.
> > I dont notice any difference in speed on my machine though. It's probably just the animation in the dev branch making it feel like it's faster.
> 
> The question was more about why we need routes at all (for model manipulation and viewing, not the about page. That makes sense as a route), not the difference between this implementation of routes and the one on dev.

I think what @aschnapp was getting at is that some pretty standard behavior (like being able to refresh while the modal is open) just breaks things instead of taking you to where you left off. We could opt out of changing the visible route in the browser at all when we open / navigate the modal but, generally, that's pretty standard and meaningful feedback for the user (and the devs) to confirm that the correct things are happening as they progress through the site. There's other little things, too, like not being able to scroll through the modal with pageup/pagedown or arrow keys.

It's also feedback that 'should' come at no cost. Like right now, if you transition between Home & About the route switch is instantaneous and you're where you want to be without any visual sign that stuff had to load. That's something we should be able to achieve with routes and something I'm sure we can figure out with a little more time.
 
More generally, though, this is a massive modal & it contains a lot of interactivity (basically 90% of the functionality of the front end) when it's generally supposed to be a very critical / quick / lightweight piece of information that interrupts whatever you're doing to inform you of something important or get confirmation of something. As an example, lets say the user spent a bunch of time filling out a form and accidentally clicks cancel. With routes we can just have a modal that pops up if changes were made asking users to confirm that they want to discard them. With a modal, what do we do? Snackbars don't block page interaction so people could still interact with the form with them up. We can't really have a modal in a modal either because that doesn't even make sense. There's no clean or idiomatic solution to the problem.

aneuhold:
@Tengelma ah okay. That makes more sense. The modal within modal confirmation dialog justifies it personally. And when a new page is used for the form entry, then the routes seem like a great next step to give some logical feedback to the user where they are.

UI should definitely be worked on to get there, but the concept as a whole makes more sense now. Thanks for that.

Tengelma (approved):
The logic seems like its at a stable point & bugs have been resolved. I'm ok with sending this through so people can start building from it.

client/src/utils/Store.js
```diff
@@ -116,14 +116,17 @@ export default class API {
...
         params: this.flatten(model),
         model_name: name,
       });
-      await Promise.all([this.setModel(name, model), this.getPlotData()]);
+      await Promise.all([this.setModel(newName, model), this.getPlotData()]);
+      await this.cloneModel(name, newName);
+      await this.deleteModel(name);
     } catch (error) {
       console.error(error);
       // better error handling here, some vue event?
```
aneuhold, line 132: This is a kind of confusing function, because it doesn't just update a model. Is it actually for updating a model or is it only for the case where a model needs to be updated AND a model needs to be renamed at the same time? Either way, the documentation on the function should indicate what it should be used for and when the `setModel` function should be used instead.

aneuhold (changes requested):
The loading bar when a model is created is much more intuitive now. 

Just one thing with how the Store should be used now and this is approved.

client/src/utils/Store.js
aschnapp: Why rename is not considered an "update" to a model? I chose to add renaming in this funtion because rename is technically an update. 
I dont understand the confusion around setmodel and updatemodel. setmodel is a local thing while update is a local *and* a server request. That's why setmodel is called in updatemodel.

aneuhold: What if you want to update the model but not rename it?

aschnapp:
Closing this in favor of #58

PR #56, opened 2021-01-25
US 195 Plot Rendering

vperuzzi:
# Summary
This PR implements User Story HM-195 Plot Rendering. Features added from this user story include:

1, Rendering of plot data for models utilizing `chart.js` and `vue-chart.js` packages.
2. Addition of `plotData` object to `Store.js` for storing computed values for plot.
3. Addition of `plotDetails` object to `Store.js` for storing plot scale, x-axis labels, and y-axis labels.
4. Addition of `getPlotData()` method to `Store.js`. This method makes a request to `/get_plot_data`.
5. Addition of `mapToChartData()` method to `Store.js`. This method maps the response from `/get_plot_data` to the format expected by `vue-chart.js`'s `renderChart()` method. **NOTE: This method should probably be refactored and moved to another file once we have refactored it into a computed property of Graph.Vue. 
6. Addition of `Chart.vue` component which extends the Line component of `vue-chart.js`.
7. Change from GET to POST for endpoint `/get_plot_data`. REST principles typically discourage associating a body with a GET request, plus axios prohibited me from sending a GET request with a body. We can discuss potentially renaming this endpoint.

image
Please note that the Latex in the axis labeled is unable to be parsed. Obviously we can simply trim it from the label, but I have left it there for now. 

## Includes
/components/Chart.vue

## Developer Checklist:

### Additional Notes
I am aware that the performance could be optimized by not repeatedly repopulating the `plotData` object with calls to `get_plot_data` for _all_ models. Due to the limitations of vue's reactivity system, unless this entire object is reset, the reactivity system won't detect changes. There are several ways to bypass these limitations this (as I'm sure you all are aware) but I ran out of time trying to implement them. We can add a user story in a future sprint for reducing the amount of models requested in `/get_plot_data` and only manipulating the data that we need.


aneuhold (approved):
This looks pretty dang good! As far as the Latex, could we just add a user story for the next sprint to add [MathJax](https://www.mathjax.org/)? Seems like a pretty standard library for rendering Latex in the browser. I use it personally with my notes for school so I know it works well, but really any library will do. That one just seems to be the most popular.

aneuhold:
Oh @vperuzzi it does need to be merged with dev still it looks like.

vperuzzi:
> 
> 
> Oh @vperuzzi it does need to be merged with dev still it looks like.

I forgot to merge, good catch. Updated.

aschnapp (approved):
It looks good!
I wonder if you could use querystrings for the GET? 
Maybe the numbers can be enhanced/simplified in the future too.

Tengelma (approved):
This is awesome!

Having get_plot_data as a POST makes sense because it calculates and caches values on the server if they haven't been computed already or something changes. We can definitely rename it to reflect that -- was an oversight on my part when I wrote it. 

I wouldn't worry too much about optimization rn since it seems pretty speedy as is. The server endpoint is already setup to let you request particular models, though, so we can definitely introduce that on the client at some point.

PR #57, opened 2021-01-25
Sprint 7

aneuhold:
The site has enough functionality that we felt making a PR into main was warranted. 😄


client/src/views/About.vue
```diff
@@ -0,0 +1,47 @@
...
+        <p>The backend for this app is written in Python. If you have a working
+        python installation you can install it simply with
+        <code>pip install halomod</code>.</p>
+
+        <p>TheHaloMod is hosted on a cluster maintained by the
+        <a href="http://loco.lab.asu.edu/">Low-frequency Cosmology</a> (LoCo)
+        lab at the <a href="http://sese.asu.edu%20target%3D/">School of Earth
+        and Space Exploration</a> at ASU. Shout-outs to Judd Bowman, Danny Jacobs,
+        Adam Beardsley and Matt
+        Kolopanis for assisting greatly with server setup and maintenance.</p>
```
steven-murray, line 25: Add yourselves here!

docs/pull_request_template.md
```diff
@@ -0,0 +1,21 @@
+### Overview
```
steven-murray, line 1: shouldn't this file be in `.github/PULL_REQUEST_TEMPLATES/`?

server/.flake8
```diff
@@ -0,0 +1,41 @@
...
+    
+max-line-length = 88
+# Should be 18.
+max-complexity = 25
+exclude =
+    development/*
+# Not ready for this yet.
+per-file-ignores =
+    scripts/hmf: F821
+    scripts/hmf-fit: F821
```
steven-murray, line 23: these can be removed

server/.flake8
```diff
@@ -0,0 +1,41 @@
...
+    attr
+    exc
+    obj
+rst-directives =
+    note
+    warning
+    versionadded
+    versionchanged
+    deprecated
+    seealso
```
steven-murray, line 41: these can most likely be removed as well

server/config.py
```diff
@@ -0,0 +1,17 @@
+import redis
+
+# flake8: noqa
+
+
+class Config:
+    # environ.get('FLASK_KEY')
+    SECRET_KEY = 'g+hH\xebiv\x9f\x8e\xb8\x95\xed?]>\x81,\x1c\\\xc1\x0fm\x95\x96'
```
steven-murray, line 8: should this be on github?

server/halomod_app/__init__.py
```diff
@@ -0,0 +1,335 @@
...
+import redis
+from flask_session import Session
+from werkzeug.exceptions import InternalServerError, HTTPException
+import sentry_sdk
+from sentry_sdk.integrations.flask import FlaskIntegration
+
+sess = Session()
+
+
+def create_app(test_config=None):
```
steven-murray, line 21: is there a good reason to have all of this in a function? I haven't written flask stuff before, but that seems like a weird design

server/halomod_app/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        # we have to set all _params whose model has been changed to {}
+        # so that they don't get carry-over parameters from other models.
+        for k, v in kwargs.items():
+            if k.endswith("model") and v != getattr(this, k).__class__.__name__:
+                this.update(**{k.replace("model", "params"): {}})
+
+        return this
+
+
+def create_canvas(objects, q: str, d: dict, plot_format: str = "png"):
```
steven-murray, line 40: this function should now be unnecessary

server/halomod_app/utils.py
```diff
@@ -0,0 +1,362 @@
...
+        "ylab": r"Mass Function $\left( \frac{dn}{d\ln M} \right) h^3 Mpc^{-3}$",
+        "yscale": "log",
+    },
+    "dndlog10m": {
+        "xlab": MLABEL,
+        "ylab": r"Mass Function $\left( \frac{dn}{d\log_{10}M} \right) h^3 Mpc^{-3}$",
+        "yscale": "log",
+    },
+    "fsigma": {
+        "xlab": MLABEL,
```
steven-murray, line 163: all of this should be somewhere in the client-side code, right?

steven-murray (approved):
Hey @halomod/thm-capstone, this is awesome. It's really hard for me to properly review the `vue` stuff. It all looks like magic to me. But having seen the product, it *looks* good :-)

One thing I didn't understand was how the model parameters for each kind of model are input -- they don't seem to appear in the `.vue` files for the forms, so I'm assuming it's something more automatic?

I'll have to get ya'll to go through some of the code with me to show me what does what soon.

client/.browserslistrc
```diff
@@ -0,0 +1,3 @@
+> 1%
+last 2 versions
+not dead
```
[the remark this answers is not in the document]
aneuhold: This describes what browsers we are targeting with the project. [See here for the documentation on the file](https://github.com/browserslist/browserslist#browserslistrc). It is parsed by most IDEs to help with flagging certain constructs as unusable for the browsers we are targeting, or it will help with code suggestions while writing the application for constructs that can be used with those browsers.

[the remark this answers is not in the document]
aneuhold: will add this as a comment. 😄

client/src/components/Graph/index.vue
```diff
@@ -0,0 +1,87 @@
...
+    };
+  },
+  components: {
+    Plot,
+    Error,
+  },
+  async created() {
+    const plotChoices = [...this.READ_ONLY.plotTypes];
+    if (plotChoices.includes('dndm')) {
+      this.plotChoice = 'dndm';
```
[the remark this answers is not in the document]
aneuhold: That is actually taken out of the most recent code that will is being created for the plot display. It was a preventative measure, in case the name of the plot changed, it would still choose the first plot that is listed.

client/src/constants/backend_constants.js
```diff
@@ -0,0 +1,759 @@
+import deepFreeze from 'deep-freeze';
+
+export default deepFreeze({
```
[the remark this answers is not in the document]
aneuhold: @Tengelma is this something we can do now?

client/src/constants/chart_data_sample.json
```diff
@@ -0,0 +1 @@
...922807578,"y":6.981793097024311e-29},{"x":10471285480508986,"y":5.199461324590415e-29},{"x":10715193052376092,"y":3.854025029129137e-29},{"x":10964781961431828,"y":2.843371190214006e-29},{"x":11220184543019652,"y":2.0884083224023794e-29},{"x":11481536214968794,"y":1.52682128208271e-29},{"x":11748975549395304,"y":1.1106082572860802e-29},{"x":12022644346174180,"y":8.034636727806735e-30},{"x":12302687708123812,"y":5.782171124052548e-30},{"x":12589254117941714,"y":4.138907597273661e-30},{"x":12882495516931322,"y":2.9472922542370935e-30},{"x":13182567385564102,"y":2.088046867544941e-30},{"x":13489628825916506,"y":1.4711799018292694e-30},{"x":13803842646028866,"y":1.0303246650587916e-30},{"x":14125375446227496,"y":7.172110647526023e-31},{"x":14454397707459280,"y":4.9625373984597685e-31},{"x":14791083881682132,"y":3.4129773251234525e-31},{"x":15135612484362072,"y":2.3336728589393477e-31},{"x":15488166189124860,"y":1.5861008724277284e-31},{"x":15848931924611108,"y":1.0709693974023266e-31},{"x":16218100973589332,"y":7.180924889707392e-32},{"x":16595869074375562,"y":4.7823296160613877e-32},{"x":16982436524617460,"y":3.162964921247002e-32},{"x":17378008287493834,"y":2.077862290340157e-32}]}]}
```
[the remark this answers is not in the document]
aneuhold: That actually isn't used anymore. It is going to be deleted. Thank you for catching that.

client/src/router/index.js
```diff
@@ -0,0 +1,104 @@
...
+    name: 'Home',
+    component: Home,
+  },
+  {
+    path: '/about',
+    name: 'About',
+    // route level code-splitting
+    // this generates a separate chunk (about.[hash].js) for this route
+    // which is lazy-loaded when the route is visited.
+    component: () => import(/* webpackChunkName: "about" */ '../views/About.vue'),
```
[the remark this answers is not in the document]
aneuhold: This was changed in [this commit](https://github.com/halomod/TheHaloMod-SPA/commit/c408d29bf6fdbdd333981791285e4995855f6a7c). That commit also has a description of why it was changed.

docs/pull_request_template.md
aneuhold: It works either way: [Github documentation on the location of the file](https://docs.github.com/en/github/building-a-strong-community/creating-a-pull-request-template-for-your-repository).

server/config.py
aneuhold: @Tengelma is it okay for this key to be here?

server/halomod_app/utils.py
aneuhold: Ended up using it again in #69 out of necessity 😕

client/src/constants/backend_constants.js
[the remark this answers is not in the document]
Tengelma: I think so? There are a lot of items in BACKEND_CONSTANTS that weren't automatically generated but if we a) move as much of that stuff as we can out and b) modify the script we used to generate this initially to get everything into an usable format, it's definitely possible. I think Cosmology is the main thing that didn't make it with the original script so that will need some work. We might even be getting to a point where we could support more generic form templates like you suggested early on @aneuhold.

I think it would be a 'one and done' sort of thing where that script runs when starting the server and saves the constants somewhere so we can grab the same object with some initial_data() endpoint. Can tinker with this next sprint?

server/config.py
Tengelma: It's mainly used to keep sessions secure. I think its OK in the near term but once we properly deploy the site we'll want to pull it from some local file that contains it.

Basically just need something in our startup shell script that says

```
if SECRET_KEY file doesn't exist:
  generate a key
  save secret key to file SECRET_KEY
set FLASK_KEY environment variable to value in file
```

and then grab that value from ENV when starting the python server. Would add that file to our .gitignore, obviously.

server/halomod_app/__init__.py
Tengelma: Haha, this refactor has been on the to-do list for a while. We will eventually have related functionality grouped in separate files and leave endpoint stuff out of __init__ so its properly just a startup script.

client/babel.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  presets: [
+    '@vue/cli-plugin-babel/preset',
+  ],
+  plugins: [
+    '@babel/plugin-proposal-optional-chaining',
+  ],
+};
```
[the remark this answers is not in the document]
Tengelma: So there's this thing called ECMAScript -- its basically a standard agreed upon by the tech overlords discussing what a general purpose 'scripting' language should be able to do. Every now and then, the standard changes. Javascript is based on that standard and changes with it. The issue is that some browsers don't support the 'new' ECMAScript standard and can't run the 'new' Javascript that came with it, so code that uses that standard might not be compatible with every browser.

Babel is a 'transpiler' that basically takes the 'new' javascript and converts into the 'old' javascript so it can run anywhere.

This config file is basically just saying we're using babel's standard (preset) transpiler that automatically detects the platform it will be running on and does all the conversions for it. That's opposed to the developer having more fine grain control over what conversions happen for what browsers. The plugin added is 'optional chaining' which Vue 2.0 (the version we're using) doesn't support out of box because its based on older ECMAScript standard than the one that includes that feature. 

Optional chaining basically gives us the freedom to test whether a layer of an object hierarchy even exists before going deeper. As an example, with the following

grandparent?.parent?.child?

we test whether grandparent exists before looking at parent and we test whether parent exists before looking at child. That makes it really easy to work with objects that don't necessarily have every layer present because you make it impossible to check for member variables of a NULL variable. It's used in a couple of places where we want to verify something in a nested hierarchy has been initialized before we allow a certain operation.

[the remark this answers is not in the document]
Tengelma: And yes, we can add some documentation to the file that explains this at a high level!

client/src/constants/backend_constants.js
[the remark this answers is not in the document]
aneuhold: @Tengelma absolutely. Sounds like a great idea generating on server start, because of course the library isn't going to change in memory while the server is running. haha. Didn't think of that at first so good pointing that out. Also looking forward to discussing this tomorrow 😄

server/halomod_app/utils.py
aneuhold: It is pulled in to the client side code from the server. In a way it is there. This would be good to discuss at the meeting tomorrow though.

client/.browserslistrc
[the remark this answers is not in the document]
steven-murray: thanks!

client/babel.config.js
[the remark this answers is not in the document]
steven-murray: Love the explanation, thanks!

client/src/components/Graph/index.vue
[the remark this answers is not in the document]
steven-murray: OK, cool.

client/src/constants/backend_constants.js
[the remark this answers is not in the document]
steven-murray: That sounds like the right plan. Things like the `backend_constants` and the initial plot data should be generated on server start.

client/src/constants/chart_data_sample.json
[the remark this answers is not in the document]
steven-murray: Sweet, OK.

server/config.py
steven-murray: Sounds good.

steven-murray: Should I make an issue to remind ourselves of this?

server/halomod_app/__init__.py
steven-murray: OK cool.

server/halomod_app/utils.py
steven-murray: Hmm, interesting. Does that mean that matplotlib is being used to generate the figure just for a pdf?

aneuhold: @steven-murray At the moment, yes that is correct.

steven-murray: OK. We can talk about this more this evening, but I wonder if we couldn't just do a hacky "convert" of the `.svg` on the client to `.pdf` and serve it up.

PR #58, opened 2021-01-28
US221 Form State Management [PARTIAL]

Tengelma:
Hi all. This is a pull request working towards the completion of US221 Form State Management. It focuses on groundwork and refactoring upstream from the forms themselves to simplify the work done on a form by form basis.

It does a lot but the big things:
- it uses Alex's routing logic as a base
- pushes our form initialization object to another file
- introduces rename functionality to the client/server (and corresponding tests on both sides)
- modifies Home UI to support renaming of models
- separates Create and Edit views
- introduces confirmation modals when saving/creating/leaving /create or /edit
- moves initial name selection logic out of the form itself and into the modal associated with creating on /create

It shouldn't affect anyone's progress on their US except for Alex because everything was so closely tied together. I am requesting reviews so we can potentially get this pulled into dev & build from it. Would consider it high priority for reviews because of how much has moved around.


aschnapp (changes requested):
I sent the things I believe that can be changed in your slack pms

Tengelma:
> I sent the things I believe that can be changed in your slack pms

Recent commits did the following:
- Resolved the navbar issue
- Brought the loading indicator in from your branch
- Adjusted file structure like you suggested

I am reluctant to combine Edit/Create views because of the following reasons:
- Content and logic are substantially different (introducing 30 ternary operators makes code less readable)
- Clearer separation of concerns with different routes having their own file
- What would we even name the view that combines edit & create and has a boatload of conditional logic in their dependent on what route you're actually in?
- I get that its not DRY but this format is being reused ONCE and I think the gains in code clarity (both in the route structure and within the file itself) outweigh the benefit of forcing all these differences into a single Vue component.

.github/workflows/client.yaml
```diff
@@ -32,12 +32,12 @@ jobs:
       with:
         # npm cache files are stored in `~/.npm` on Linux/macOS
         path: ~/.npm
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}-1
```
aneuhold, line 35: Just out of curiosity, how did this addition of `-1` fix it?

client/src/constants/forms.js
```diff
@@ -0,0 +1,107 @@
...
+    model: 'mass-definition',
+    title: 'Mass Definition',
+    id: 'mass-definition',
+  },
+  {
+    component: Concentration,
+    model: 'tracer-concentration',
+    title: 'Tracer Concentration',
+    id: 'tracer-concentration',
+    props: {
```
aneuhold, line 26: Excellent. So much more clean.

client/src/utils/Store.js
```diff
@@ -130,6 +130,31 @@ export default class API {
...
+      ]);
+      this.state.models[newName] = model;
+      delete this.state.models[oldName];
+      this.state.modelNames = this.getModelNames();
+      this.getPlotData();
+    } catch (error) {
+      console.log(error);
+    }
+  }
+
```
aneuhold, line 157: Woot!

aneuhold (changes requested):
The buttons on the bottom left and the confirmation dialog when creating a new model look great! Love that addition 🌈. In general, the model naming, and creation process is awesome now. Great work on all that. 

Only a couple questions before approving, and also, the spacing for the models is quite large now. Does it need to be that large or can we narrow that down a bit? See below:

image

This looks like it was a ton of work. Awesome job to you and Michael for getting this all together. 

P.S. Someone got their formatter to work! lol

.github/workflows/client.yaml
Tengelma: Ah, it didn't end up fixing it. Github's caching is a bit weird because you can't manually clear it. I kept hitting the same node_modules cache (that wasn't working because of some replace all I did at some point) and it led to repeated build failures so I was just trying to get a clean slate. 

Eventual solution was just to npm update locally, generate a new package-lock.json and ship that up instead. Thank you Stack Overflow !

Tengelma:
@aneuhold There's no great reason for the models being as spaced as they are. It's mainly just because what's displayed is an md-field and md-input under the hood and I didn't think to change it. I'll see if I can update the formatting soon.

.github/workflows/client.yaml
aneuhold: Awesome. Can the key be changed back then really quick?

Tengelma:
@aneuhold Your concerns should be resolved. Lemme know if there's anything else to do.

aneuhold (approved):
Amazing! Looks fantastic 😄

aschnapp:
See Forms.vue in views
Just pushed an additional file that does not have any ternaries to review. Your changes are still there. If we decide to use yours, just delete this file before we merge. 

I am reluctant to combine Edit/Create views because of the following reasons:
- Content and logic are substantially different (introducing 30 ternary operators makes code less readable)
No ternaries
- Clearer separation of concerns with different routes having their own file
The largest part of it is a v-if
- What would we even name the view that combines edit & create and has a boatload of conditional logic in their dependent on what route you're actually in?
Naming should not be a reason to keep us from keeping it DRY

- I get that its not DRY but this format is being reused ONCE and I think the gains in code clarity (both in the route structure and within the file itself) outweigh the benefit of forcing all these differences into a single Vue component. 
We are going to be adding error reporting. Do we really want to be copying code between the two documents on every single change just to keep it in sync? You already had to add a loading indicator in the exact same place for both files, how many more are we going to be doing?

aschnapp (approved):

PR #59, opened 2021-01-30
US216 Latex Rendering

aneuhold:
## Overview
This PR implements US 216, Latex Rendering. It includes some big overhauls to how the charting is rendered on the front-end, which was discussed with Tia before-hand. But it does not touch the backend end-point. So that is the same.

### Charting Framework Change and Latex Rendering Package

Some key issues were stumbled upon when trying to render Latex. Those will be laid out below in order, to help understand why this change seemed to need to be made:

* Chart.js only allows string data types to be provided for labels on the x and y axis
* [Katex](https://katex.org/) can only render to two things: an HTML element, and HTML string. 
* Chart.js only renders with [Canvas](https://www.w3schools.com/html/html5_canvas.asp). This has some side-effects vs [SVG](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/svg). When an element is placed on a canvas, it cannot be manipulated after the fact. It can not be queried either. It is like using Microsoft Paint vs using PowerPoint to make a graphic. Canvas is like paint, and SVG is like PowerPoint in that in an SVG you can manipulate the items individually after being placed.
* Because Chart.js chart components cannot be manipulated afterwards, the x axis and y axis labels cannot be directly swapped with rendered Latex, or have the rendered latex provided to Chart.js to render.
* Manually trying to place rendered Latex over the canvas raised issues with sizing, adjusting the chart if the latex was large, and generally look and feel problems.
* [D3.js](https://d3js.org/) is another charting framework. it has much more granular control over many aspects of the process. 
* D3 creates charts by building an SVG, meaning that it is entirely configurable, and rendered items can be placed where labels typically are in a clean way.
* D3 provides all of the functionality of Chart.js
* As stated previously, Katex can only export HTML. HTML cannot be embedded in an SVG in an efficient way which still allows re-scaling and transforms (rotations and position changes).
* [MathJax](https://www.mathjax.org/) can export to SVG, which blends well with the charting and positioning.
* MathJax allows very specific components to be retrieved and used, which cuts down dramatically on the bundle-size. So that was done to make it still efficient.

Therefore, D3, and, specifically TeX input and SVG output of MathJax, was chosen.

For some examples of the final product, see below:

image

image

image

## Includes:
- client/package-lock.json 
- client/package.json
- client/public/index.html 
- client/src/components/Chart.vue 
- client/src/components/Graph/D3Plot.vue 
- client/src/components/Graph.vue → client/src/components/Graph/index.vue 
- client/src/components/SubmitButton.vue 
- client/src/utils/Store.js 
- client/src/views/Home.vue 

## Developer Checklist:

## Notes:
- Some known bugs that will be handled in [US 217](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-217) are:
   - Resizing the window doesn't automatically update the chart. Selecting a new chart does though
   - When a chart with large numbers on the x-axis is chosen, they overlap
- MathJax has to be pulled in through a script tag unfortunately. But this was handled pretty easily by putting that in `index.html`. It appends a globally accessible `window.MathJax` object which, so far, seems to work incredibly fast.
- MathJax delimiters do not work for some reason. Spent a good 3 hours on that alone, and ended up building a quick home-brew delimiter parser for Latex. It works quick and does it's job. That is located in `D3Plot.vue` under the `processLatexString` function.

## Refs:
- [US 216](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-216)


vperuzzi (approved):
Anton, this looks amazing. The D3 Plot component works and looks great. Runs perfectly and LaTeX is correctly rendered for all models. Approved! 💯

client/src/components/Graph/D3Plot.vue
```diff
@@ -0,0 +1,307 @@
+<template>
+  <div id="d3-chart" class="plot"></div>
+</template>
+<script>
+import * as d3 from 'd3';
```
Tengelma, line 5: Anyway we could have a different alias? Would be great if we didn't see 'd3' anywhere in the code beyond the imports -- could let the language match the use case whereas the name of the library tells us nothing about what it does.

client/src/components/Graph/D3Plot.vue
```diff
@@ -0,0 +1,307 @@
...
+          .tickFormat(''));
+
+      // add the Y gridlines
+      svg.append('g')
+        .attr('class', 'grid')
+        .attr('transform', `translate(${leftPadding},0)`)
+        .call(make_y_gridlines()
+          .tickSize(-(w - rightPadding - leftPadding))
+          .tickFormat(''));
+    },
```
Tengelma, line 143: Can this logic be moved to another file since its just JS and doesn't use Vue functionality?

client/src/components/Graph/D3Plot.vue
```diff
@@ -0,0 +1,307 @@
...
+        .split(/\$/g)
+        .reduce((label, string, index) => {
+          if (index % 2 === 0) {
+            return `${label}\\textrm{${string}}`;
+          }
+          return label + string;
+        // Uses an empty string as the second argument to reduce so that it uses
+        // that as the starting point instead of the first array value.
+        }, '');
+    },
```
Tengelma, line 164: We'll also probably need latex rendering in other places so this might make more sense in utils?

client/src/components/Graph/D3Plot.vue
```diff
@@ -0,0 +1,307 @@
...
+      svg.select('#legendColor')
+        .call(colorLegend);
+      const legendNode = svg.select('#legendColor').node();
+      const legendBBox = legendNode.getBoundingClientRect();
+      const targetX = w - legendBBox.width;
+      const targetY = (h / 2) - (legendBBox.height / 2);
+      legendNode.setAttribute('transform', `translate(${targetX},${targetY})`);
+      return legendBBox.width;
+    },
+  },
```
Tengelma, line 285: Literally all of this logic would be be perfectly fine to put in a pure JS class to keep it separate from the data/component itself. Would prevent you needing the thin wrapper for the vue component you have at index.js.

Tengelma (changes requested):
Overall great work. Sucks that you had to use a completely different library to get everything working with latex but the end result is solid.

Most of my comments are related to code quality / clarity -- the big thing is just moving all the pure JS stuff to another file to separate library interaction / logic from the presentation a bit more. The D3Graph component has no local data and basically nothing in the template  -- no reason for it to be a component.

Once this stuff is resolved, you've got my approval!

client/src/components/Graph/D3Plot.vue
aneuhold: What would be the solution then? `d3` is short and does explain what library is being used when the term is entered. It is used a lot too.

aneuhold: Also many libraries don't explain what it does. For example: `axios`.

client/src/components/Graph/D3Plot.vue
aneuhold: It does use Vue functionality in that it is called whenever the data changes from the store. You are right it could be moved to a JS file separately, with an argument which specifies the id of the element to modify, although what would the file be called and where should it go? All of this code is very much having to do with the interface, so it doesn't seem like it would belong in the `utils` folder. Up to you.

client/src/components/Graph/D3Plot.vue
aneuhold: Yes this could go in the utils folder. Maybe a file called `stringUtils`? Not sure where it would go in there or what it would be named.

client/src/components/Graph/D3Plot.vue
Tengelma: It uses Vue functionality that could just as easily be one line  (watch: READ_ONLY.plotdata{buildPlot();) in the parent component. And yes its related to the interface but none of these functions interact with the components state, props, or any other functionality related to it actually being in a component. They take data as an input then build a javascript object and tie that to an html element with a provided id. 

And I'd say a function like buildPlot would very much qualify as a utility. Maybe utils/plot.js?

client/src/components/Graph/D3Plot.vue
Tengelma: string-utils.js would work!

client/src/components/Graph/D3Plot.vue
Tengelma: Fair enough on the 'axios' point -- I guess my main gripe is less with using library functions as d3.* and more with including the name of the library in the wrappers you have for it. Like 'D3Plot' as a component name, 'd3-plot' as an id name and d3plotData as an item in our store.

It's like if we created a thin wrapper for axios requests to make calling them a little easier and then proceeded to call the file axios-requests.js, the class in it AxiosRequests, and any time we had data we wanted to use as params called it axiosRequestData. It just ties the functionality / naming to the implementation instead of the content & context.

Does that make more sense?

client/src/components/Graph/D3Plot.vue
Tengelma: Also just while we're at it, it might make more sense to have the 'getPlotTypes' call in the store and call it ONCE on store initialization (and save it in store as plotTypes). Wouldn't be a big change but would clean up index.js substantially.

Tengelma:
Also just while we're at it, it might make more sense to have the 'getPlotTypes' call in the store and call it ONCE on store initialization (and save it in store as plotTypes). Wouldn't be a big change but would clean up index.js substantially.

client/src/components/Graph/D3Plot.vue
aneuhold: Ah that makes sense. Definitely. Will change the file name.

aneuhold: And the other instances that it is referenced in that way.

aneuhold: Sorry. One more comment. it was done that way because while developing this, it was done side-by-side with the chart.js implementation. So that is why it stuck around. Just didn't think about changing it until you mentioned it here. Thank you.

client/src/components/Graph/D3Plot.vue
aneuhold: This was done as requested.

client/src/components/Graph/D3Plot.vue
aneuhold: This was done as requested.

client/src/components/Graph/D3Plot.vue
aneuhold: This was done as requested.

PR #60, opened 2021-02-02
Us221 form state management update

aschnapp:
## Overview
This is part of US221 that changes the default form data to user defined form data in the edit component.
Also merges create/ and edit/ into forms.vue.

## Includes:
New file: views/Forms.vue
Every form has been changed, unused ones deleted.
Other places may be changed too, please see files tab.

## Developer Checklist:
- [x] The code follows the Quality Policy file on Google Drive
- [] My code has been developer-tested and includes unit tests
- [x] I have considered proper use of exceptions
- [x] I have eliminated IDE warnings

## Notes:
When reviewing, please double check that each form component is working correctly, and that state is properly cloned. There's a lot, I hope I didnt miss anything. 

## Refs:
- [US 216](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-221)

PR #62, opened 2021-02-06
US220 T232 Speeding Up Transitions 

Tengelma:
#60 is a prerequisite.

This PR introduces Vue's keep-alive functionality to reuse the forms component between route switches. This modifies every part of the form hierarchy to reinitialize values from props whenever 'activated', the lifecycle hook associated with a return to a cached component, occurs. The initial render of forms is still as slow as it was before but everything after that is instantaneous because the component never has to be recreated/mounted.

Developer Checklist:
- [] My code has been developer tested and includes unit tests

Notes:
Concentration form has an outstanding bug where the values in local state (and the values being displayed, as a result) do not match the values passed in through props. They instead match whatever the PREVIOUS set of props was. I have tinkered with this a bit but given how much time I've spent overhauling forms, recently, I'd really like to pawn this one out to someone else. Anyone want to give it a go?


aneuhold (approved):
Rock on. This does make it go quite a bit faster. 🚀

aschnapp (approved):
Activated is interesting. Very fast :)

PR #64, opened 2021-02-08
US 196 deploy server

dblevin1:
### Overview
User Story 196, server deployment. This PR includes all the required Docker files to be self contained. I also wrote a helpful python script to pass the appropriate environment variables depending on the passed parameters, this is documented in the README. A serverDeployement configuration file is included that works uses github actions to ssh into the server and download the latest repo (this has not been tested yet, we need Steven to put the username and password into github). 

### Includes:
- .github/workflows/serverDeployment.yaml
- server/Dockerfile
- server/README.md
- server/config.py
- server/docker-compose-dev.yml
- server/docker-compose-prod.yml
- server/docker-compose.yml
- server/dockerStart.py
- server/nginx/default.conf
- server/run.sh
- server/settings/base.env
- server/settings/dev.env
- server/settings/prod.env

### Developer Checklist:
- [x] My code has been developer-tested and includes unit tests.

### Notes:
- ~~Steven can add his USERNAME and PASSWORD into github secrets, I created the secrets already he will just have to go and update them.~~
- ~~Once we get access to his server we can test deployment and enable the docker-compose lines.~~

### Refs:


aschnapp (changes requested):
basically stuff I asked during the meeting today. 

no need for dev container, no need for base then, and probably dont need nginx, env can be part of docker compose

server/dockerStart.py
```diff
@@ -0,0 +1,66 @@
+import sys
```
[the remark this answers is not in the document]
dblevin1: Yes this sets up some of the ENV variables that the user can define (host, port, production or not). It also starts the container, ends the container, and cleans up when exited.
Of course its possible to start the containers without this file by exporting the ENV variables, but for future ease of use I think it is necessary.

For now I'm just going to comment out the nginx stuff till we get it live, in case it is needed.
I changed the .env extension to .conf files to avoid confusion. using these files and passing them to the docker-compose is the "docker way" of doing it. It is possible to export the FLASK_ENV and FLASK_DEBUG vars in dockerStart.py thus not needing the conf files, but its standard practice to use the files.

If we end up not needing nginx, docker-compose-prod.yml can be deleted.

server/README.md
```diff
@@ -1,13 +1,61 @@
...
+export FLASK_RUN_HOST=0.0.0.0
+export FLASK_RUN_PORT=5000
+docker-compose up --build --remove-orphans
+```
+
+To stop the container:
+
+`docker-compose down`
+
+## Manual Start
```
aschnapp, line 38: It should be called "development" or something along those lines (please find a name better than "development" lol)

server/run.sh
```diff
@@ -13,8 +13,14 @@ do
     "--test" )
       python -m pytest;;
     "--dev" )
+      redis-server
       export FLASK_APP=halomod_app
       export FLASK_ENV=development
       flask run;;
+    "--prod" )
```
aschnapp, line 20: is this called somewhere? 
We dont need this for ourselves

server/dockerStart.py
```diff
@@ -0,0 +1,62 @@
...
+import os
+
+
+def printHelp():
+    print("USAGE: dockerStart.py [OPTIONS]\n" +
+          "  Running dockerStart.py with no options is equivalent to:\n" +
+          "  dockerStart.py -b --port=5000 --url=0.0.0.0\n\nOptions:" +
+          "     -h, --help       Show this command list\n" +
+          "(Turns on debug mode for Flask, and installs pytest)\n" +
+          "     -p, --prod       For production (Enables nginx)\n" +
```
aschnapp, line 13: Docker should be prod only. We do not need to specify prod or not

server/docker-compose-prod.yml
```diff
@@ -0,0 +1,42 @@
+version: '3.5'
```
aschnapp: Why do we need to split docker composes? 
Isnt this prod only?

.github/workflows/serverDeployment.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - name: DeployServer
+      uses: appleboy/ssh-action@master
+      with:
+        debug: true
+        host: ${{ secrets.HOST }}
+        username: ${{ secrets.USERNAME }}
+        key: ${{ secrets.SERVERKEY }}
+        port: ${{ secrets.PORT }}
+        script:  |
+          export ENV_FILE=settings/base.conf
```
aschnapp, line 30: isnt this a duplicate of dockerStart.py?

server/settings/base.conf
```diff
@@ -0,0 +1,3 @@
+COMPOSE_FILE=docker-compose.yml
```
aschnapp, line 1: we don't need this file, docker is prod only, and run.sh has this already.

server/settings/prod.conf
```diff
@@ -0,0 +1,3 @@
+COMPOSE_FILE=docker-compose-prod.yml
```
aschnapp: (optional) this can be hardcoded or saved inside dockerStart.py. There's no need for an additional file.

aschnapp (changes requested):

server/dockerStart.py
dblevin1: I removed the dev option and removed this script because the user will not need to specify a environment.

.github/workflows/serverDeployment.yaml
dblevin1: No this is the Github action that Github is running on the the ssh connection to the server.

server/run.sh
dblevin1: Removed.

server/settings/base.conf
dblevin1: Removed.

server/README.md
```diff
@@ -4,10 +4,13 @@
...
-  - [How to start the server locally](#how-to-start-the-server-locally)
-  - [Development commands](#development-commands)
-- [Architecture](#architecture)
+- [TheHaloMod-SPA Server](#thehalomod-spa-server)
+  - [Table of Contents](#table-of-contents)
+  - [Usage](#usage)
+    - [How to start the server locally](#how-to-start-the-server-locally)
+    - [Development commands](#development-commands)
+    - [Production deployment using docker](#production-deployment-using-docker)
+  - [Architecture](#architecture)
```
aneuhold, line 13: The top level `TheHaloMod-SPA Server` probably doesn't need to be included in the table of contents, or table of contents itself. But that is a small thing and won't block this from being approved.

server/README.md
```diff
@@ -28,6 +31,21 @@ Before doing any of the commands below, make sure to install the Python packages
...
+export FLASK_RUN_HOST=0.0.0.0
+export FLASK_RUN_PORT=5000
+export FLASK_KEY=${{ secrets.FLASK_KEY }}     [optional, if none is defined a default value is used]
+docker-compose up --build --remove-orphans -d
+```
+
+To stop the container:
+
+`docker-compose down`
+
```
aneuhold, line 48: Awesome 😄

server/run.sh
```diff
@@ -13,6 +13,7 @@ do
     "--test" )
       python -m pytest;;
     "--dev" )
+      redis-server
```
aneuhold, line 16: Didn't know that this could be done. It seems to work fine, running the redis-server and the python server at the same time, but it looks to block the output of the python server. Is there a way to fix this? 

Also if this is being done, the documentation for running the python server locally will need to be changed.

aneuhold (changes requested):
Just one change request for the `run.sh` file or more of an explanation.

server/run.sh
dblevin1: Good catch I didn't realize this was changed when I removed the production environment. I think it makes more sense to remove this.

server/README.md
dblevin1: Agreed

aneuhold (approved):
Looks good!

PR #69, opened 2021-02-15
US4 Multiple Format Downloads

aneuhold:
## Overview
This PR implements User Story 4, multiple format downloads. 

### UI Update for new Button

The home page UI was updated to include a new download section. It has a drop down with the different download options and a download button to execute the download. 

image

The view automatically adjusts in a mobile layout to make sense, by putting the download button at the very bottom:

image

### Plot Download

The plot download as a PDF was exceedingly difficult to pull off. Getting the SVG to render in a PDF with all of the MathJax and other fonts and styles of the page turned out to be something that needed to be coded in it's own right. As in, CSS, and JavaScript needed to be written specifically for the PDF and only for the PDF for it to work, and even then it wasn't a guarantee. So, instead, the functionality the project used to have was used to get a plot image from the server. 

### ASCII Download

This was generally copied over from Steven's work with a few changes to fit the API with Flask.

### Params Download

This was very straightforward, just state that we already had, downloaded into a JSON file. 

## Includes:
- client/package-lock.json 
- client/src/components/Download.vue 
- client/src/utils/Store.js 
- client/src/utils/plot.js 
- client/src/views/Home.vue 
- client/yarn.lock 

## Developer Checklist:

## Notes:
- Some unused imports were removed from the server side. 

## Refs:
- [US 4](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-4)


aschnapp:
@aneuhold I recommend that you use a ui button next to the + and then a modal with the dropdown for data downloads. The download card gets pushed down when you have a lot of models. 
Screen Shot 2021-02-15 at 4 02 39 PM

aschnapp:
If you put it with the toolbar, you can make it disappear like with the delete all trash can when there are no models.

aneuhold:
@aschnapp it seemed more intuitive to leave it out of the model interface because it doesn't just have to do with the models, it also has to do with the plot. If the button was in the models, intuitively it seems like it would be used to only download model data or params data, but there is an option to download an image of the plot as well.

aneuhold:
@aschnapp also the goal isn't to have it visible in one viewport, the page can be scrolled 😄

aschnapp:
Ok that makes sense. I'll approve it. Make sure to fix the conflict though.

aschnapp (approved):

client/src/utils/Store.js
```diff
@@ -195,6 +216,24 @@ export default class API {
...
+  getModels = async () => {
+    const modelNames = this.getModelNames();
+    const modelPromises = modelNames.map((modelName) => this?.state.models[modelName]);
+    const allModelObjs = clonedeep(await Promise.all(modelPromises));
+    const allModels = {};
+    modelNames.forEach((modelName, index) => {
+      allModels[modelName] = allModelObjs[index];
+    });
+    return allModels;
+  };
```
Tengelma, line 235: Would be good if this was called getAllModels just to be more explicit.

Can you explain what the foreach is doing here and why you can't just return allModelObjs? Maybe included as comments inline? Assuming its related to models being stored as promises in the current setup of the store but would be good to have documentation for it just for transparency.

Tengelma (changes requested):
Just some little changes suggested. Works well and does what its supposed to do.

I will say I'm kind of on board with Michael's suggestion to have a button instead of an entire panel but also agree that, as the UI is currently layed out, that doesn't _quite_ work. Given that everything on home is so strongly coupled / grouped in functionality, it might make sense to rework the UI a bit so the components don't seem so disjoint. Maybe something we could discuss at the next sprint planning?

client/src/utils/Store.js
aneuhold: @Tengelma having promises in the store is exactly why. Will add a comment about it now.

Tengelma (approved):

PR #70, opened 2021-02-15
HM235 add cosmo tests

aschnapp:
## Overview
Adds cosmo tests. Only one file has been changed.

## Includes:
- CosmologyForm.spec.js

## Developer Checklist:

## Refs:
- [US 4](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-235)

### DIFF
--- a/client/tests/unit/CosmologyForm.spec.js
+++ b/client/tests/unit/CosmologyForm.spec.js
@@ -54,7 +54,7 @@ describe('Mounted CosmologyForm', () => {
     }
   });
 
-  test('emits onChange event whenever model is changed', async () => {
+  test('emits onChange event whenever model selection is changed', async () => {
     const emitted = wrapper.emitted();
     let prevCount = 0;
     for (const option of options) {
@@ -68,7 +68,7 @@ describe('Mounted CosmologyForm', () => {
     }
   });
 
-  test('emits onChange event whenever model subparams is changed', async () => {
+  test('emits onChange event whenever model core params are changed', async () => {
     const emitted = wrapper.emitted();
     const { model } = wrapper.vm;
     let prevCount = 0;
--- a/client/tests/unit/CosmologyForm.spec.js
+++ b/client/tests/unit/CosmologyForm.spec.js
@@ -22,6 +22,20 @@ describe('Mounted CosmologyForm', () => {
     expect(wrapper.vm.model.cosmo_model).toBe('Planck15');
   });
 
+  test('changes model values when values are changed', async () => {
+    const model = wrapper.vm.$data.model;
+    Object.keys(model).forEach(async (key) => {
+      if (typeof model[key] === 'number') {
+        const o = wrapper.vm.model[key];
+        model[key] += 1;
+        await localVue.nextTick();
+        await localVue.nextTick();
+        const n = wrapper.vm.model[key];
+        expect(n).toBe(o+1);
+      }
+    })
+  });
+
   test('changes model parameters when model is changed', async () => {
     for (const option of options) {
       if (wrapper.vm.model.cosmo_model === option) continue;
--- a/client/tests/unit/CosmologyForm.spec.js
+++ b/client/tests/unit/CosmologyForm.spec.js
@@ -75,7 +75,6 @@ describe('Mounted CosmologyForm', () => {
     let prevCount = emitted.updateCosmo.length;
     const params = Object.keys(wrapper.vm.model.cosmo_params);
     for (const param of params) {
-      const val = wrapper.vm.$data.model.cosmo_params[param];
       wrapper.vm.$data.model.cosmo_params[param] += 0.1;
       await localVue.nextTick();
       await localVue.nextTick();
--- a/client/tests/unit/CosmologyForm.spec.js
+++ b/client/tests/unit/CosmologyForm.spec.js
@@ -31,6 +31,7 @@ describe('Mounted CosmologyForm', () => {
       await localVue.nextTick();
       const newParams = wrapper.vm.model.cosmo_params;
       expect(oldParams).not.toBe(newParams);
+      console.log(option);
     }
   });
 
@@ -40,30 +41,16 @@ describe('Mounted CosmologyForm', () => {
       await localVue.nextTick();
       await localVue.nextTick();
       const params = Object.keys(BACKEND_CONSTANTS.Cosmo_params[option].cosmo_params);
+      const htmlStrings = {
+        H0: 'H<sub>0</sub>',
+        Ob0: 'Ω<sub>b</sub>',
+        Om0: 'Ω<sub>m</sub>',
+        z: 'Redshift',
+        n: 'n<sub>s</sub>',
+        sigma_8: 'σ<sub>8</sub>',
+      }
       for (const param of params) {
-        switch (param) {
-          case 'H0':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*H<sub>0</sub>.*')));
-            break;
-          case 'Ob0':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*Ω<sub>b</sub>.*')));
-            break;
-          case 'Om0':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*Ω<sub>m</sub>.*')));
-            break;
-          case 'z':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*Redshift.*')));
-            break;
-          case 'n':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*n<sub>s</sub>.*')));
-            break;
-          case 'sigma_8':
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*σ<sub>8</sub>.*')));
-            break;
-          default:
-            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp(`.*${param}.*`)));
-            break;
-        }
+        expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp(`.*${htmlStrings[param] || param}.*`)));
       }
     }
   });
--- a/client/tests/unit/CosmologyForm.spec.js
+++ b/client/tests/unit/CosmologyForm.spec.js
@@ -23,17 +23,17 @@ describe('Mounted CosmologyForm', () => {
   });
 
   test('changes model values when values are changed', async () => {
-    const model = wrapper.vm.$data.model;
-    Object.keys(model).forEach(async (key) => {
+    const { model } = wrapper.vm.$data;
+    for (const key of Object.keys(model)) {
       if (typeof model[key] === 'number') {
         const o = wrapper.vm.model[key];
         model[key] += 1;
         await localVue.nextTick();
         await localVue.nextTick();
         const n = wrapper.vm.model[key];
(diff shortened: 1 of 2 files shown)


client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,108 @@
...
+          case 'n':
+            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*n<sub>s</sub>.*')));
+            break;
+          case 'sigma_8':
+            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp('.*σ<sub>8</sub>.*')));
+            break;
+          default:
+            expect(wrapper.html()).toEqual(expect.stringMatching(new RegExp(`.*${param}.*`)));
+            break;
+        }
```
Tengelma, line 66: Not DRY. 0 reason for the switch. Make a list of the tags and loop through that instead of params -- the 6 you have cases for are the only ones that appear for this model.

['H<sub>0</sub>', '*Ω<sub>b</sub>', etc...]

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,108 @@
...
+    for (const option of options) {
+      if (wrapper.vm.model.cosmo_model === option) continue;
+      const oldParams = wrapper.vm.model.cosmo_params;
+      wrapper.vm.$data.model.cosmo_model = option;
+      await localVue.nextTick();
+      await localVue.nextTick();
+      const newParams = wrapper.vm.model.cosmo_params;
+      expect(oldParams).not.toBe(newParams);
+    }
+  });
```
Tengelma, line 35: Cosmo is weird because core parameters (things outside cosmo_params) change with model changes, too. Should have tests to reflect this.

Tengelma (changes requested):
Cosmo has a lot of special cases so copy paste & replace all won't work, unfortunately. Big thing that's different is that it has three core parameters and three sub parameters and the tests you have only properly address the latter. HMFForm.spec.js is probably a better base than the one you pulled these from because the form it tests has a similar data structure.

Changes needed:
- @aneuhold saved field values into a local object 'allCosmoData' when switching between models. This is a feature on all of the forms he wrote. Should have tests to verify that those saved values don't persist between uses of the form (switching between create -> edit, etc...). Once these tests are written once we can reuse the pattern for the other forms like this
- Modifying the model modifies the value of the core parameters that are outside of cosmo_model so you need checks to ensure that a) they're still around and b) their values have been updated in a way that's consistent with what we'd expect
- You have a test to see if modifying things in cosmo_params emits an event but no equivalent test for the other half of the params

Hope this all makes sense. Let me know if you have any questions.

client/tests/unit/CosmologyForm.spec.js
aschnapp: What core parameters?

If core parameters change, model changes, therefore. There's no need to test very highly coupled parts, because we may want to change that, but not break the tests. 
Testing that this component returns the right output given the inputs should be sufficient.

aschnapp:
@Tengelma 
> Changes needed:
> 
>     * @aneuhold saved field values into a local object 'allCosmoData' when switching between models. This is a feature on all of the forms he wrote. Should have tests to verify that those saved values don't persist between uses of the form (switching between create -> edit, etc...). Once these tests are written once we can reuse the pattern for the other forms like this
allCosmoData is a "state" (like in react) it should reset whenever it's mounted. Unless we're testing vue itself I do not see a need to test that. 
> 
>     * Modifying the model modifies the value of the core parameters that are outside of cosmo_model so you need checks to ensure that a) they're still around and b) their values have been updated in a way that's consistent with what we'd expect
Are you talking about core and sub?
> 
>     * You have a test to see if modifying things in cosmo_params emits an event but no equivalent test for the other half of the params
I'll modify that to test the entirety of the model. There's no need to test params alone.

Tengelma:
> @Tengelma
> 
> > Changes needed:
> > ```
> > * @aneuhold saved field values into a local object 'allCosmoData' when switching between models. This is a feature on all of the forms he wrote. Should have tests to verify that those saved values don't persist between uses of the form (switching between create -> edit, etc...). Once these tests are written once we can reuse the pattern for the other forms like this
> > ```
> 
> allCosmoData is a "state" (like in react) it should reset whenever it's mounted. Unless we're testing vue itself I do not see a need to test that.
> 
> > ```
> > * Modifying the model modifies the value of the core parameters that are outside of cosmo_model so you need checks to ensure that a) they're still around and b) their values have been updated in a way that's consistent with what we'd expect
> > ```
> 
> Are you talking about core and sub?
> 
> > ```
> > * You have a test to see if modifying things in cosmo_params emits an event but no equivalent test for the other half of the params
> > ```
> 
> I'll modify that to test the entirety of the model. There's no need to test params alone.

1. Forms are only mounted once now (and never reinitialized) because of how keep-alive works, hence the need to reinitialize in activated & verify that it has been reinitialized when switching between routes/contexts.
2. There are three parameters (z, m, sigma) that aren't in cosmo_params. They're updated manually in watch: 'model.cosmo_model'. That's what I'm referring to when I say 'core parameters'. Testing that they're updated correctly when a model changes is basically verifying that the updateOptions function does what its supposed to (and continues to do what its supposed to).
3.  The reasoning behind testing individual parameters is because a) it's easy enough to do and b) the model itself is stable but how we interact with it isn't, necessarily. If someone refactors this and deletes a line in the process or uses the wrong variable name somewhere then a test would fail and force them to address it instead of just being a silent error that takes a year to see because who's going to notice that 1 param isn't updating properly.

client/tests/unit/CosmologyForm.spec.js
Tengelma: It's testing our logic within the watch function not whether the watch triggers. I agree that we don't need to test Vue features but we do need to test whether the code we wrote using Vue features produces expected results.

In the same vein, it's not testing whether or not about testing whether 'updated' triggers, its testing whether or not the correct thing happens _when_ updated is triggered.

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    const params = Object.keys(wrapper.vm.model.cosmo_params);
+    for (const param of params) {
+      const val = wrapper.vm.$data.model.cosmo_params[param];
+      if (typeof val === 'string') {
+        if (val === 'full') {
+          wrapper.vm.$data.model.cosmo_params[param] = 'relaxed';
+        } else {
+          wrapper.vm.$data.model.cosmo_params[param] = 'full';
+        }
+      } else {
```
aneuhold, line 86: It looks like just this section be removed and it will work well.

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,95 @@
...
+  test('changes model parameters when model is changed', async () => {
+    for (const option of options) {
+      if (wrapper.vm.model.cosmo_model === option) continue;
+      const oldParams = wrapper.vm.model.cosmo_params;
+      wrapper.vm.$data.model.cosmo_model = option;
+      await localVue.nextTick();
+      await localVue.nextTick();
+      const newParams = wrapper.vm.model.cosmo_params;
+      expect(oldParams).not.toBe(newParams);
+      console.log(option);
```
aneuhold, line 34: This looks to log things to the console still at the moment.

aneuhold (changes requested):
Just a couple changes and approved 😄

client/tests/unit/CosmologyForm.spec.js
aschnapp: Done

aschnapp:
@Tengelma the route changes are not going to be tested right now. The changes do work though.

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,99 @@
...
+      if (typeof model[key] === 'number') {
+        const o = wrapper.vm.model[key];
+        model[key] += 1;
+        await localVue.nextTick();
+        await localVue.nextTick();
+        const n = wrapper.vm.model[key];
+        expect(n).toBe(o+1);
+      }
+    })
+  });
```
Tengelma, line 37: This isn't testing anything meaningful. You're modifying an object and then verifying that the object was modified.

Tengelma (changes requested):
Alex, this still has the same issues I pointed out in my last review :(

Again, you have three 'core' parameters and three 'sub' parameters. You have tests in place for the 'sub' parameters but not the 'core' parameters. There needs to be checks to:
- verify changes to core parameters lead to an event being emitted
- verify changes to the cosmo_model change the value of the 'core' parameters

Like I've said before, HMFForm tests give a pretty explicit template for how both of these can be done. Feel free to copy paste and adjust as needed but please include them in this PR.

client/tests/unit/CosmologyForm.spec.js
aschnapp: A couple lines down there's the exact same test that tests params. They're not the same object. Line 26 is $data.model Line 33 is model itself, with things changing in data.
This tests if vmodel does anything

There's another test that tests if things get emitted when vmodel changes a couple functions below.

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -80,6 +80,17 @@ describe('Mounted CosmologyForm', () => {
...
+      if (typeof model[key] === 'number') {
+        prevCount = emitted.updateCosmo.length;
+        wrapper.vm.$data.model[key] += 1;
+        await localVue.nextTick();
+        await localVue.nextTick();
+        expect(emitted.updateCosmo.length).toBeGreaterThan(prevCount);
+        prevCount = emitted.updateCosmo.length;
+      }
+    }
   });
```
Tengelma: It would be good if this was in a separate test just to further hone in on where things might fail.

client/tests/unit/CosmologyForm.spec.js
Tengelma: wrapper.vm.$data.model == wrapper.vm.model. They're the same object (reference and all). vue-test-utils gives you getters and setters at the vm level so you don't have to go through $data to get to the object. 

The test you have modifies one of the model objects keys and sees if that 
The test that follows changes cosmo_model and sees if the sub parameters have updated.
The test I'm asking for changes cosmo_model and sees if the core parameters have update.

aschnapp:
@Tengelma what sub parameters?

aneuhold (approved):
Updates look good!

client/tests/unit/CosmologyForm.spec.js
```diff
@@ -0,0 +1,101 @@
...
+      prevCount = emitted.updateCosmo.length;
+      wrapper.vm.$data.model.cosmo_model = option;
+      await localVue.nextTick();
+      await localVue.nextTick();
+      expect(emitted.updateCosmo.length).toBeGreaterThan(prevCount);
+      prevCount = emitted.updateCosmo.length;
+    }
+  });
+
+  test('emits onChange event whenever model subparams is changed', async () => {
```
Tengelma, line 71: Since these are outside cosmo_params this should be 'whenever model main/core/whatever you want to call it' params are changed.

Name of the test after this should have a corresponding change to indicate that its focused on the stuff inside cosmo_params.

Tengelma (approved):
I am approving but please don't merge until you fix the two little things I suggested (just naming stuff). Once those are good to go this looks good.

PR #71, opened 2021-02-17
US249 Forms Integration Testing (PARTIAL)

Tengelma:
### Overview
This implements a handful of tests to verify state is flowing correctly through the entire form pipeline. There's still more to do on the US but these tests are ready so may as well get them into dev.

Tests included are:
- a check to ensure that modifications to any subform will cause an update to the top-level view's state (current form data flows up)
- a check to ensure that modifications to the top level view's state will NOT modify subform state (current form data doesn't flow down)
- a check to ensure sub-objects of the top level view's state are not being used directly by individual subforms (reference check to make sure data is only flowing through event system)
- a few tests to verify that each part of the form hierarchy is rendering properly

### Included Files
- Minor modifications to every form (non-breaking)
- A small update to setup.js to avoid VueObserveVisibility console warnings while testing
- Integration.spec.js

### Developer Checklist
- [X] My code has been developer tested and includes unit tests


vperuzzi (approved):

aneuhold:
What is the `subform_id` doing specifically? Also, as far as the props to each component, would it be okay to specify their type or document them a bit? Steven specifically asked for that in the latest comments he made so it might be helpful in this instance.

Tengelma:
> What is the `subform_id` doing specifically? Also, as far as the props to each component, would it be okay to specify their type or document them a bit? Steven specifically asked for that in the latest comments he made so it might be helpful in this instance.

@aneuhold subform_id helps us tie the the particular form to its section of the composite form data. It's not used in the client main program flow but I used it during testing, here, so that I could look at an individual subform's state and say 'hey is this the same thing as what my parent has or not'? It's identical to the 'id' passed in to the component BUT id is kind of a keyword in vue & becomes the 'id' of the underlying div instead of a prop we can access from the context of the subform.
 
image

Tengelma:
I could have kept that data contained to tests but I think it's an useful thing to have as a prop as we move more towards dynamic generation in places that we can. The id of the subform is also the prefix of the _model and the _params in the vast majority of forms so its an easy way to build the correct output (kinda like we do in concentration/profile form already).

As for documentation, I think it's fine if we add documentation to everything but think its kind of outside the context of this user story. There isn't any significant functionality / changes beyond the tests so going through and saying the same thing for every subform_id seems excessive right now.

client/tests/setup.js
```diff
@@ -4,3 +4,11 @@ import VueObserveVisibility from 'vue-observe-visibility';
...
 Vue.use(VueMaterial);
 Vue.use(VueObserveVisibility);
+
+global.console = {
+  warn: jest.fn(),
+  log: console.log,
+  info: console.info,
+  debug: console.debug,
+  error: console.error,
+};
```
aneuhold, line 14: What does this do? Might be helpful to have a comment above it.

aneuhold (changes requested):
Looks pretty good. Once the one question is answered and the merge conflict is solved it is approved :)

aneuhold:
@Tengelma just checking in. It looks like my review was requested and the recent commit says that `setup.js` has documentation but it doesn't look to have documentation on this end. Did something not go through?

Tengelma:
> @Tengelma just checking in. It looks like my review was requested and the recent commit says that `setup.js` has documentation but it doesn't look to have documentation on this end. Did something not go through?

Looks like it. I will get this sorted soon. Sorry for the mishap.

aneuhold:
@Tengelma no worries 🙂

Tengelma:
> @Tengelma no worries slightly_smiling_face

@aneuhold Should be good to go now. I had the comments locally but git wasn't registering them as a modification to the file? Was using /* comment */ syntax before and not sure why that didn't work but the // comments did.

aneuhold (approved):

PR #72, opened 2021-02-20
US234 Axis Drop Down

aneuhold:
### Overview
This PR implements US 234, which was the multiple axis drop-down US. What it does right now, is allow each of the plots as defined in the `KEYMAP` in the server to be combined with their listed x-axis. This opens the door to allow any x-axis to be combined with any y-axis. Although, without better understanding how HMF takes that information and generates the plot data, it will have to wait until then. It does make it very easy to implement in the future though once that is better understood. 

Image of the drop-downs:

image

Image of the options for each:

### Includes:
- client/.browserslistrc 
- client/jsconfig.json 
- client/src/components/Graph/Plot.vue 
- client/src/components/Graph/index.vue 
- client/src/components/Models/index.vue 
- client/src/components/forms/index.vue 
- client/src/constants/chart_data_sample.json 
- client/src/constants/plotNames.js 
- client/src/router/index.js 
- client/src/utils/Store.js 
- client/src/utils/latex.js 
- client/src/utils/plot.js 
- client/tests/example_data/plotTypes.json 
- docs/pull_request_template.md 
- server/halomod_app/utils.py 

### Developer Checklist:

### Notes:
- Comment was added to `.browserlistrc` to describe its use.
- `jsonconfig.json` was added to help our IDEs understand the Vue `@` prefix notation on imports and properly import types and information from the associated file.
- The x-axis issue where it wouldn't always use a logarithmic x-axis was fixed
- A simple `setError` function was added to the store to allow for quick error reporting to the user
- `chart_data_sample` was removed because it wasn't being used anywhere in the code
- `plotNames` was added as a translation for the different axis names. This was pulled from the original `TheHaloMod`.
- The code splitting was removed in `router/index.js` because it didn't seem like it was being used.
- Latex string processing was moved into it's own file so that it can be used easily elsewhere in the application to make rendered latex SVGs.
- `pull_request_template` was updated with what we discussed at the meeting
- Small linting error was fixed in `utils.py`

### Refs:
- [US 234](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-234)


aneuhold:
This is leaving out the changes that Steven recently requested because they can be implemented in the new PR for the more dynamic ability of the x and y axis. But it can be used as a basis for that branch to help save time.

server/halomod_app/__init__.py
```diff
@@ -127,10 +127,16 @@ def get_names():
...
+    #   plotOptions: <data from KEYMAP in utils.py>
+    # }
     @app.route('/get_plot_types', methods=["GET"])
     def get_plot_types():
-        res = utils.KEYMAP
+        res = {
+            'xLabels': utils.XLABELS,
+            'plotOptions': utils.KEYMAP
+        }
         return jsonify(res)  # returns full key map of plot types
```
Tengelma, line 140: I think its safe enough to move these constants to the front end and scrap this endpoint. utils.XLABELS and utils.KEYMAP were part of the old halomod web app implementation which is not going to be maintained or updated anymore. As a result, these will be stable and there shouldn't be a need to pull them from the server dynamically. If steven wants to update them he'll need to do it manually regardless of whether its on the client or the server.

Would you be open to having the download functionality on the client send X & Y labels with the request? Then you can keep it JUST on the client and not have to worry about it during startup anymore!

docs/pull_request_template.md
```diff
@@ -12,6 +12,7 @@ When a code reviewer has a more complete understanding of how and why a particul
 - [ ] My code has been developer-tested and includes unit tests
 - [ ] I have considered proper use of exceptions
 - [ ] I have eliminated IDE warnings
+- [ ] I have included tasks associated with this pull request in the Sprint Progress Sheet.
```
Tengelma, line 15: Good to see this!

client/src/utils/latex.js
```diff
@@ -0,0 +1,23 @@
...
+  MathJax.texReset();
+
+  const correctedString = processLatexString(inputLatex);
+
+  // Use the first child because the parent has unneeded extra fluff
+  return MathJax.tex2svg(correctedString)
+    .firstChild;
+};
+
+export default createLatexSvgFromString;
```
Tengelma, line 23: Nice! Maybe in keeping with library tendencies you can just call this function latex2svg? A little more concise.

client/src/router/index.js
```diff
@@ -58,23 +58,17 @@ const routes = [
   {
     path: '/about',
     name: 'About',
-    // route level code-splitting
-    // this generates a separate chunk (about.[hash].js) for this route
-    // which is lazy-loaded when the route is visited.
-    component: () => import(/* webpackChunkName: "about" */ '../views/About.vue'),
```
Tengelma, line 64: Probably want to revert these (and maybe name the chunks in accordance with the routes while you're at it)

Tengelma (changes requested):
A handful of pretty small suggestions. Works great just would be nice to simplify things in some places.

Looking forward to seeing how the new setup will work once you get what Steven wanted working :)

server/halomod_app/__init__.py
aneuhold: It sounds like these things should be included in the NEW US. Like we had discussed at the meeting: https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-278

client/src/router/index.js
aneuhold: This branch doesn't have anything to do with routes or naming the chunks so it will just be reverted.

client/src/utils/latex.js
aneuhold: Ok

aneuhold:
> A handful of pretty small suggestions. Works great just would be nice to simplify things in some places.
> 
> Looking forward to seeing how the new setup will work once you get what Steven wanted working :)

@Tengelma this branch will not include the new things that Steven requested at the most recent meeting. Those are being included in the new US like we had discussed at the meeting.

server/halomod_app/__init__.py
Tengelma: Unresolving so @aschnapp can see. I think this is fair game for either US because this is used by both approaches to axis selection.

The basic point is that utils.KEYMAP and utils.XLABEL are more tied to the display side of things and not the halomod side of things. I think it makes sense to have these constants on the client-side, instead, as a result. If we get it there it also simplifies the back and forth quite a bit because you just need to include the X and Y axis options in the request to the server's getPlot and don't have to duplicate data on the client and server. It also removes the need for:
- getPlotTypes in store
- getPlotTypes on server
- the need to store constants (plot types) in our store
- etc...

I think this is a fairly simple refactor with decent payoff in code clarity so if it doesn't make it on this PR I will definitely want it on Alex's. Maybe you two can figure out who should be responsible for it since its so closely tied to both user stories?

Tengelma (approved):
Approving but see comments.

aneuhold:
https://github.com/halomod/TheHaloMod-SPA/pull/72#discussion_r582373761

@Tengelma that should definitely be changed by @aschnapp . Just trying to get this in so he has something to base his off of and not work as much. If we are both still actively working on the exact same thing it doesn't make any sense.

Tengelma:
> [#72 (comment)](https://github.com/halomod/TheHaloMod-SPA/pull/72#discussion_r582373761)
> 
> @Tengelma that should definitely be changed by @aschnapp . Just trying to get this in so he has something to base his off of and not work as much. If we are both still actively working on the exact same thing it doesn't make any sense.

Yep, I already approved so if that's what y'all settle on OK. I just want to make sure it makes it in somewhere :)

client/src/utils/plot.js
```diff
@@ -1,7 +1,7 @@
 import * as d3 from 'd3';
 import Debug from 'debug';
 import { legendColor } from 'd3-svg-legend';
-import processLatexString from './stringUtils';
+import createLatexSvgFromString from './latex';
```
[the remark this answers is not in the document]
aneuhold: Had a feeling you might bring that up. `latex2svg` is actually the name of a very popular function in a library elsewhere that has to do with. I used it at one point. So while it is fine to go with your suggestion in the other location, it doesn't feel right here, because it was used here at one point. The other function name makes more sense personally. https://www.npmjs.com/package/latex2svg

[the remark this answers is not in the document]
aneuhold: Feel free to rename it in one of your commits, but I disagree with the renaming in this location.

aschnapp:
#72 

@Tengelma I can take over from here.

aschnapp (approved):

PR #76, opened 2021-02-26
US281 Server Constants Export

aneuhold:
## Overview
This PR implements User Story 281, exporting constants from the server side. It does this by using an altered version of Steven's original script for creating the backend constants, and it additionally gathers the cosmo constants.  

### ~~Route-specific files~~

~~Some of the routes were refactored into their own files to help with separation of concerns and organization. Everything still passes tests and works the same. From this perspective it seems to be easier to read as well for those. See [blueprints](https://flask.palletsprojects.com/en/1.1.x/blueprints/) on how that works for Flask. But basically a blueprint works the same as a flask `app` object.~~

### Cosmo options were found!

Finally the source of the different cosmo models was found (Planck13, Planck15) etc. `astropy.cosmology.parameters`. Two of them don't seem to work with HMF yet though. So those are skipped. 

### Library updates

Some libraries were updated to allow for the constants to be pulled. All tests still pass and the site works the same.

### Skipped constants

Because some constants like `n`, `z`, and `sigma_8` do not have different values for each cosmological model, they were left out of the `cosmo_defaults` section of the constants from the server. They DO have different values as hard-coded by Steven on the front-end of the old site. Because of this, it seems like a better idea to leave constants that cannot be pulled from libraries to be implemented on the front-end, that way they all stay in one place. 

### Open to suggestions

If there are any constants that are missed, that might be included in the libraries, please feel free to speak up and I will try to grab those as well. 

## Includes:
- server/.flake8 
- server/halomod_app/routes/ascii.py 
- server/halomod_app/routes/constants.py 
- server/halomod_app/routes/create.py 
- server/halomod_app/routes/plot_types.py 
- server/halomod_app/utils.py 
- server/requirements.txt 
- server/tests/test_ascii.py 
- server/tests/test_constants.py 
- server/tests/test_create.py 
- server/tests/test_plot_types.py 

## Developer Checklist:

## Refs:
- [User Story 281](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-281)


server/halomod_app/routes/constants.py
```diff
@@ -0,0 +1,95 @@
...
+        cosmo_model = hmf.cosmo.Cosmology(cosmo_model=getattr(hmf.cosmo, choice))
+        cosmo_defaults.setdefault(choice, {
+            "H0": cosmo_model.cosmo.H0.value,
+            "Ob0": cosmo_model.cosmo.Ob0,
+            "Om0": cosmo_model.cosmo.Om0
+        })
+
+    backend_constants.setdefault('cosmo_defaults', cosmo_defaults)
+
+    return backend_constants
```
Tengelma, line 80: It would be good if there was a little more guidance, here, on what the output of each of these for loops will look like, structurally. I think it would help establish a more fine grain correspondence between our current backend_constants and the code that produced it, which makes modifying it down the line easier.

server/halomod_app/routes/plot_types.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+# This endpoint returns the details of all the different plot types that
+# can be used to represent a halo model.
+#
+# expects: None
+# outputs: {
+#   xLabels: <data from utils.py>,
+#   plotOptions: <data from KEYMAP in utils.py>
+# }
```
Tengelma, line 14: Maybe make this a docstring for consistency?

server/halomod_app/__init__.py
```diff
@@ -148,6 +108,7 @@ def get_plot_types():
     #              <model_label>: {"xs": <array_of_xs>, "ys": <array_of_ys>},
     #              ...
     #           }}
+
     @app.route('/get_plot_data', methods=["POST"])
     def get_plot_data():
```
Tengelma, line 114: Not super clear on why some endpoints were moved to blueprints and others weren't?
Same with conversion of comments to docstrings.

I get that it wasn't in the original scope of this user story but the inconsistency makes the server-side code a little harder to follow and navigate. Not really sure what the solution is except completing the abstraction or moving the abstraction to another US / branch.

Tengelma (changes requested):
I have very little to say about the code that exports constants but...

There are ALOT of changes here that were outside the scope of the User Story and those changes were applied pretty inconsistently. I am on board with using blueprints and breaking up our server into smaller pieces but maybe there's room for more discussion and thought on how that should actually be done. This refactor is probably something someone needs to commit to as a large-ish user story that overhauls how the entire API is structured, alongside the client-side changes that accompany that.

With that in mind, I think these changes should be excluded from this branch and moved elsewhere until it can be discussed a little more and be done in a more comprehensive way. It's a great start and we can definitely use & extend what you've got if you can tuck the changes away on another branch but I would like to see this particular pull request simplified to what the acceptance criteria called for.

Does that make sense? Can obviously discuss if you have any questions.

server/halomod_app/routes/plot_types.py
aneuhold: Oh, must have missed that. Will do.

server/halomod_app/__init__.py
aneuhold: The goal was to test the waters. Seems like it did it's job. If it needed to be reverted, then it wouldn't be too hard, and if it needed to be completed, still wouldn't be too hard. See other comment below for a more thorough explanation.

aneuhold:
> I have very little to say about the code that exports constants but...
> 
> There are ALOT of changes here that were outside the scope of the User Story and those changes were applied pretty inconsistently. I am on board with using blueprints and breaking up our server into smaller pieces but maybe there's room for more discussion and thought on how that should actually be done. This refactor is probably something someone needs to commit to as a large-ish user story that overhauls how the entire API is structured, alongside the client-side changes that accompany that.
> 
> With that in mind, I think these changes should be excluded from this branch and moved elsewhere until it can be discussed a little more and be done in a more comprehensive way. It's a great start and we can definitely use & extend what you've got if you can tuck the changes away on another branch but I would like to see this particular pull request simplified to what the acceptance criteria called for.
> 
> Does that make sense? Can obviously discuss if you have any questions/thoughts.

There wasn't a lot about this that was outside of the scope of the user story, some, yes, but not a lot. The separate routing with blueprints made a lot more sense because of all the initialization logic that was exclusive to the `constants` endpoint. Putting all of that in the main thread of server startup would have been a mess and probably been rejected by the group because it was already very complicated in there and Steven specifically requested that it was reduced in some way. The libraries had to be updated in order for this to work, and the cosmo-options are part of the constants. The parts that were out of scope were the other routes which were converted to their own files and the separation of a single function into Utils, so that could be accomplished. What else was there?

The files were separated into their own routes along with `constants` as an example to test the waters. If it caused a problem, then it could be reverted easily, if it needed to be completed, then it could be completed easily. Please note that the conversion didn't require a change in tests whatsoever and the logic is exactly the same. Meaning there was literally no functional change to the endpoints that were moved into their own file. 

@Tengelma with that said, would you like the changes to be removed and have just constants in it's own file (or added to the main server startup if wanted, seems like a lot to throw in there, but up to you), or have the routes completed? It is easy to do either way. We discussed this change at the previous meeting so it seemed like something that should just be done whenever someone had the chance. Kind of like adding documentation. It is a simple refactor after all.

server/halomod_app/routes/constants.py
aneuhold: Agreed. It was so dang complicated was having a hard time figuring out what was happening. Figured Steven would know because he wrote it, but you are absolutely right. WE need to know how this works lol.

server/halomod_app/routes/plot_types.py
aneuhold: That one has been fixed 😄

Tengelma:
@aneuhold The main concern I have is that splitting things up in this way (1:1 route to file) and storing everything in a flat routes directory isn't really a huge improvement. Now file names are ambiguous and you have to go into the file to see what part of the application it interacts with. As an example, I have no idea what the 'create' file does until I look at it and say 'oh this is for models'. I have no idea what 'constants' does until I go in there and see 'oh this is just the defaults for all the different models'. The current single file format at least groups things by functionality and provides pretty clear documentation of what it does right next to the name of thing.

Imo, we probably want to break things up more by functionality --
model/<operation> for crud endpoints
download/<format> for download endpoints
constants/<category> for constants endpoints
etc... 
and that WOULD require changes to the tests on this end & changes to our front-end store to accommodate.

If we do do that, we'd also probably want to break utils up by those same contexts (makes it easier to extend) and keep __init__ to a single 'blueprint' import & pure server initialization logic.


So I think there's a lot more work to do to do it right and it makes more sense to me to keep this PR relatively clean and focus on the acceptance criteria. Extending our utils file is fine and changes to documentation are welcome but I think route stuff should be relocated to another branch and demoed at our next Sprint planning. That work for you?

aneuhold:
@Tengelma it seems like we are going to use blueprints regardless and split things up into separate files regardless, so this would be a good start. But oh well. Everything into one file it is. Working on it now.

aneuhold:
@Tengelma just as a side note though, going back to a single file does make it very hard to reason about the code at the moment. So please start working on the refactor you are looking for soon. It is almost prohibitively difficult to work on more complex endpoints without some kind of allowance for abstraction or separation of concerns.

Tengelma:
> @Tengelma it seems like we are going to use blueprints regardless and split things up into separate files regardless, so this would be a good start. But oh well. Everything into one file it is. Working on it now.

I agree that its a good start and something we can build from which is why I'm suggesting you move these changes to another branch so we can still do that. Still outside the context of the User Story and still not enough thought or consistency put into to warrant inclusion in dev, imo. But that's just, like, my opinion, man. If you want to convince two other teammates to review and send it through, feel free to do so :man_shrugging:

Tengelma:
> @Tengelma just as a side note though, going back to a single file does make it very hard to reason about the code at the moment. So please start working on the refactor you are looking for soon. It is almost prohibitively difficult to work on more complex endpoints without some kind of allowance for abstraction or separation of concerns.

I can appreciate that the server is in need of a refactor and that it would help to get it done sooner rather than later. Could be something someone can tackle early next Sprint?

And I am not super opinionated on the specifics of how things should be structured but just want the pattern used to be something that's understandable and stable long-term. For real, if you have other good ideas on how that can be done or feedback on what I proposed, I'd love to hear it!

server/halomod_app/routes/constants.py
Tengelma: We can ask Steven to walk us through it next meeting haha

Tengelma:
BTW, if you move the generate_constants function to utils (which seems fine as a home for it given the stuff currently in there), __init__.py only needs a few extra lines, which avoids a lot of your concerns about adding more complexity to that file while we figure routes out :sweat_smile:

server/halomod_app/routes/constants.py
aneuhold: This is done.

aneuhold:
@Tengelma it was reverted back to a single file per your request.

Tengelma (approved):
This looks good. Thank you for updating things @aneuhold.

server/halomod_app/__init__.py
```diff
@@ -409,4 +395,16 @@ def ascii():
 
         return send_file(buff, as_attachment=True, attachment_filename="all_plots.zip")
 
+    backend_constants = utils.generate_constants()
+
+    @app.route('/constants', methods=["GET"], strict_slashes=False)
```
aschnapp, line 400: So why do we need a specific endpoint for this? Why can't it be generated as part of the deployment/run process? 
Only the client needs it so we can create a script that's called inside package.json that generates backend_constants and saves it in the correct location.

aschnapp (changes requested):

server/halomod_app/__init__.py
aneuhold: Would this run on server start?

aneuhold: How would it access the python packages on the client side?

aschnapp: No run on the CI, using an sh file called from npm or elsewhere

PR #81, opened 2021-03-02
US279 T297 Tests for Basic GenericForm

Tengelma:
### Overview
This is a pull request to finish off US279, which was focused on getting a GenericForm template going to simplify long-term maintenance of our forms.

It includes eight tests, which all used the 'Bias' form's metadata as a base:
- check to verify component data is initialized correctly
- check to verify cache is updated correctly when model selection changes
- check to verify available parameters change when model selection changes
- check to verify an onChange event is emitted whenever model selection or params have changed
- check to verify that component data reflects changes in its initial_state prop
- check to verify that the component renders the correct number of fields when model selection changes
- check to verify that correct values and names for each field are rendered in the resultant HTML
- check to verify that the correct title is rendered in the resultant HTML

### Includes:
- Tests are in tests/unit/GenericForm.spec.js
- jest.config.js was adjusted to except the GenericForm tests in the tests it currently excludes
- @/constants/forms.js' top level data structure was changed from an array to an object to help with testing
- @/components/forms/index.js had minor changes to support the above

### Developer Checklist:

### Notes:
- This is being merged into GenericForms again so that it doesn't interfere with the current user experience of dev.
- Why just the Bias form context? Hypothetically, since this component is being used for EVERY basic form, verifying that ONE of them works correctly is enough. I chose Bias because of familiarity but this choice is arbitrary.
- I did try to make more generic tests work (kind of like integration.spec.js does on dev) but left it behind as the amount of hardcoded sample data required to make it work ballooned. This is probably an user story in and of itself and seems beyond the scope of this one.
- The current tests don't check for the exceptions (checkbox for one value of a particular model of our profile forms and a dropdown for one value for a particular model of concentration form). I think these are high effort & low value at this point and since GenericForm is likely to change a lot next Sprint as we include the more advanced forms it makes sense to me to hold off on it for now.

### DIFF
--- a/client/jest.config.js
+++ b/client/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   preset: '@vue/cli-plugin-unit-jest',
   testPathIgnorePatterns: [
-    '<rootDir>/tests/unit/.*Form.spec.js',
+    '<rootDir>/tests/unit/(?!Generic).*Form.spec.js',
   ],
   moduleFileExtensions: ['js', 'json', 'vue'],
   transform: {
--- a/client/src/components/forms/index.vue
+++ b/client/src/components/forms/index.vue
@@ -6,7 +6,7 @@
       class="md-primary"
       md-fixed>
       <md-list>
-        <md-list-item v-for="(form, index) in forms" :key="index"
+        <md-list-item v-for="(form, index) in Object.values(forms)" :key="index"
           :class="{'router-link-active': currentlyVisible == form.title}"
           :to="`#${form.id}`">
           {{form.title}}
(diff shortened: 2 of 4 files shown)


aneuhold (approved):

aschnapp (changes requested):
Feel free to ignore if this is not something relevant. 
It seems like the forms are incomplete? where's cosmology? 
This is what I'm getting:
Screen Shot 2021-03-08 at 5 09 06 PM

aschnapp (approved):
Actually remove my previous comment, this isnt going into dev. 

Approving now and merging

PR #82, opened 2021-03-05
US270 Fix autoscroll on refresh functionality for forms

Tengelma:
### Overview
Simple but hacky fix for the forms view not scrolling to the correct subform on refresh / manual address input. It forces scroll on the updated lifecycle hook (which happens during a page reload) in the FormView component. Even though created / mounted might sound more 'correct', neither works in this situation.  

### Includes:
- Minor updates to Forms.vue

### Developer Checklist:
- 
### Notes:
- There is a lot of chatter online about Vue's scrollBehavior not working properly when used as the docs suggest. It SHOULD be scrolling on refresh/reload and it SHOULD let you use selectors instead of interacting with the DOM directly, but neither of these actually works as expected. Not sure what the deal is but I guess this approach is functional and unobtrusive enough :slightly_frowning_face:

### DIFF
--- a/client/src/views/Forms.vue
+++ b/client/src/views/Forms.vue
@@ -113,7 +113,21 @@ export default {
     }
     this.current = clonedeep(this.initial);
   },
+  updated() {
+    this.scrollToAnchor();
+  },
   methods: {
+    scrollToAnchor() {
+      this.$nextTick(() => {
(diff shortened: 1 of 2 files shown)


aneuhold (approved):
It works! Great fix 😄

aschnapp (approved):

PR #83, opened 2021-03-06
HM278 axes

aschnapp:
## Update: 
(Latest) Removes /plot endpoint and download functionality in favor of @vperuzzi's upcoming changes. 
#### (old)
Tests are fixed. 

### Overview
Expands potential x axes types. They are all mlabel though. 
Every x axis that Steven requested is included with the exception of the ones that are not currently in KEYMAP (n(<m) and peak height). 

### Includes:
Please see files tab. Quite a bit has changed.
Created a new /globals folder in project root to share data between server and client. We can use that as a model for backend constants. (@aneuhold)

### Developer Checklist:

### Notes:
- Please review this asap, I will be writing tests but I want to catch errors and fix them before I get too far into the tests. 
- The branch is for a specific task, I dont want to rename it because of contribution report links
- We'll need to comb through __init__.py to get rid of all references to utils.KEYMAP and delete KEYMAP in favor of plotData.json
### Refs:
- [HM278](https://the-halo-mod.atlassian.net/browse/HM-278)


globals/plotData.json
```diff
@@ -0,0 +1,256 @@
...
+  "corr_1h_auto_tracer": {
+    "label": "1-halo tracer correlation, $\\xi_{1h}(r)$",
+    "scale": "log",
+    "name": "1-halo tracer-tracer correlation function"
+  },
+  "corr_1h_cross_tracer_matter": {
+    "label": "1-halo matter-tracer correlation, $\\xi_{1h}^{m\\times T}(r)$",
+    "scale": "linear",
+    "name": "1-halo matter-tracer correlation function"
+  },
```
aneuhold, line 26: How will the tests access these? As well as the deployed front-end and backend. This might work locally, but when it is running in CI, normally it is isolated from anything higher than the folder that is used to deploy the server/front-end.

server/halomod_app/__init__.py
```diff
@@ -323,17 +296,17 @@ def clear():
...
+        y = request_json["y"]
         img_type = request_json["img_type"]
 
         if 'models' in session:
             models = pickle.loads(session["models"])
         else:
             models = {}
-
-        # generates figure/plot
+        # generates figure/plot not working!
```
aneuhold, line 307: This does actually seem to be working locally 😛

server/halomod_app/utils.py
```diff
@@ -38,67 +39,68 @@ def hmf_driver(cls=TracerHaloModel, previous: Union[None, TracerHaloModel] = Non
...
-    for x, label in XLABELS.items():
-        if KEYMAP[q]["xlab"] == label:
-            break
-    else:
-        raise ValueError(f"The quantity {q} is not found in KEYMAP")
+    # for x, label in XLABELS.items():
+    #     if KEYMAP[q]["xlab"] == label:
+    #         break
+    # else:
+    #     raise ValueError(f"The quantity {y_key} is not found in KEYMAP")
```
aneuhold, line 64: This can be removed if it is no longer being used.

client/src/utils/Store.js
```diff
@@ -362,13 +345,16 @@ export default class Store {
...
-    if (newPlotType !== this.state.plotType) {
-      this.state.plotType = newPlotType;
-      await this.getPlotData();
+  setPlotType = async (plotType, axis, refresh) => {
+    if (plotType !== this.state[axis]) {
+      this.state[axis] = plotType;
+      if (refresh) await this.getPlotData();
     }
   }
 }
```
aneuhold, line 360: Why is `this.state.axis` not defined in the initialization part of the `Store` if it is used here?

aneuhold (changes requested):
This looks great. 

Don't get me wrong, any time that we can unify configuration files instead of having them in two different places, I am 100% for it. But the only issue seems to be the use of globals across the sever and the client. It isn't clear if that is going to work when deployed and during CI. Normally there are rules there that prevent that kind of behavior from occurring for security reasons. But, if you do get it to work in both situations, then 100% for it and going to learn from your implementation so I can use it in personal projects 😁.

aneuhold:
Also, @aschnapp we might want to leave Steven out of review requests until we have an end product on the dev branch. Otherwise it feels like it is using up his time when the rest of us can comb through it first.

aschnapp:
@aneuhold I know. It was a very specific request to make sure the graphs are correct. It wasnt for a full review.

client/src/utils/Store.js
aschnapp: axis is a string variable. either "x" or "y". It's written in the docs

server/halomod_app/utils.py
aschnapp: I want to add something to replace that, but it works at the moment and I dont have much time to get that done for this sprint.
A small enhancement for potential error if someone calls the api directly.

globals/plotData.json
aschnapp: For webpack stuff, it's actually ok, it'll be "compiled in"
Idk how python stuff will do though. The docker compose or docker files will need to copy it in.

aschnapp: The pytests and all works right now. I'm not sure about deployment though..

aschnapp: @dblevin1

aschnapp:
Tests have been fixed :)

server/halomod_app/utils.py
aneuhold: Okay can a comment be put there for that then? Right now it just looks like commented out code and no one else looking at it could really tell that someone is planning on putting something there.

client/src/utils/Store.js
aneuhold: Oh there it is. Alright cool.

server/halomod_app/__init__.py
```diff
@@ -191,14 +166,14 @@ def get_plot_data():
...
-                print(f"Error encountered getting {fig_type} for model {name}")
+                    400, f"Error encountered getting {y_param} for model {name}. {str(e)}.")
+                print(f"Error encountered getting {y_param} for model {name}")
                 print(e)
 
             res["plot_data"][name] = data  # put data in response object
 
         # put figure metadata into response
-        res["plot_details"] = utils.KEYMAP[fig_type]
+        res["plot_details"] = utils.KEYMAP[y_param]
```
aneuhold, line 176: Can all references to KEYMAP be changed now? This might make a big impact on tests, because this will show if the references to the globals files is actually working in all cases on the Python side. 

Also if KEYMAP is no longer needed, then it can probably be deleted so we don't have the same data in two places.

server/halomod_app/utils.py
```diff
@@ -38,67 +40,70 @@ def hmf_driver(cls=TracerHaloModel, previous: Union[None, TracerHaloModel] = Non
         return this
 
 
-def create_canvas(objects, q: str, d: dict, plot_format: str = "png"):
+def create_canvas(objects, x_key: str, y_key: str, plot_format: str = "png"):
+    KEYMAP = load_json('../globals/plotData.json')
```
aneuhold, line 44: It might be good to change the name from KEYMAP to something else to help differentiate. Like `PLOT_DATA` or something along those lines.

server/run.sh
```diff
@@ -10,6 +10,8 @@ do
       python3 -m pip install -r requirements.txt;;
     "--lint" )
       flake8 tests halomod_app;;
+    "--fix" )
```
aneuhold, line 13: This looks like it is a lint fix. `fix` seems very vague and Steven / future maintainers might not know what is meant here. Maybe `--lint-fix`?

aneuhold (changes requested):

globals/plotData.json
aneuhold: Okay it is looking like it is coming along then. Just a few more comments below 😄

globals/plotTypes.json
```diff
@@ -0,0 +1,83 @@
...
+      "corr_2h_auto_tracer",
+      "corr_2h_cross_tracer_matter",
+      "corr_auto_matter",
+      "corr_auto_tracer",
+      "corr_cross_tracer_matter",
+      "corr_linear_mm",
+      "sd_bias_correction"
+    ]
+  }
+}
```
Tengelma, line 83: So just to be clear this bit isn't fully implemented yet? Each of these x values should correspond to everything in the y, ya?

globals/plotTypes.json
```diff
@@ -0,0 +1,83 @@
...
+      "ngtm",
+      "radii",
+      "rho_gtm",
+      "rho_ltm",
+      "satellite_occupation",
+      "sigma",
+      "total_occupation",
+      "tracer_cmz_relation",
+      "k_hm"
+    ]
```
Tengelma, line 37: How did you decide what should / shouldn't be in x vs y?

Tengelma (changes requested):
A few general notes & questions that don't really have associated code:

It's important to make filenames & variable names consistent with what we would expect and have used in the past.
- If you're importing something from a JSON file and intend for it to be constant, we've typically used all caps & underscores (see how backend_constants, initial_state, etc are imported)
- names of files with constants should be snake case (this is even true of files, generally, in both javascript & python)

Beyond that plot_data and plot_constants aren't super descriptive of what's actually in there (this is an issue we have with a lot of our constants files but may as well catch this one here).
- I think it would be good to rename plotData to plot_axis_metadata
- I think it would be good to rename plotTypes to plot_axis_options

A bigger picture change (and something I'd suggest) would be shipping axis labels (and any other plot related metadata) with the createPlot endpoint instead of importing the display constants on client and server. It avoids the need for extra startup logic for our docker container, provides a clearer separation of concerns and makes the createPlot endpoint more general, which is never a bad thing.

Looks like this was a lot of work! Thanks for grinding it out, Alex.

Tengelma:
@aschnapp just a heads up, client-side plot downloads are working now so you don't need to worry too much about the python side of things, here. @vperuzzi should be opening a PR sometime tomorrow with the fix included

That also means all constants can be held on client without too much trouble :smile:.No need for globals after all haha

server/halomod_app/__init__.py
aschnapp: Well kindof... 
other endpoints are still using it. I did not want to refactor those as I had no idea what they did.

server/run.sh
aschnapp: I wanted it to look like sh run.sh --lint --fix (kinda like the npm one). What do you think?

globals/plotTypes.json
aschnapp: It is. All of the new ones that Steven requested are m labels. 
And yes, you have x vals should corrospond to everything in y.

globals/plotTypes.json
aschnapp: I didn't. It's Steven's list.

server/halomod_app/utils.py
aschnapp: This part has been removed in favor of @vperuzzi's changes

server/halomod_app/__init__.py
aschnapp: this is removed now.

server/halomod_app/utils.py
aschnapp: this has been removed

server/run.sh
aneuhold: Oh that definitely makes sense. Can that be put in the readme?

client/src/components/Graph/Plot.vue
```diff
@@ -39,7 +39,7 @@ export default {
   methods: {
     generatePlot() {
       debug('plot being built');
-      buildPlot(this.plotElementId, this.plotData);
+      buildPlot(this.plotElementId, this.$store.state.plot);
     },
   },
   beforeDestroy() {
```
aneuhold, line 45: `plotData` on this component is no longer needed if it isn't being used here 😄

client/src/utils/plot.js
```diff
@@ -117,10 +119,11 @@ function generateAxisLabels(svg, plotData) {
  *
  * @param {string} elementId the ID of the element to manipulate which will
  * become the parent of the SVG plot
- * @param {} plotData the plot data which should be held in state of the
- * `$store` of the Vue instance
+ * @param {} plot the plot data which should be held in state of the
+ * `$plot` of the Vue instance
  */
```
aneuhold, line 124: `$store` didn't change did it?

aneuhold (approved):
Just two things, but approved besides that. 

Much cleaner!

server/run.sh
aneuhold: Never-mind. Saw the most recent change for this 😄

Tengelma (approved):
Green light from me. Thanks for changes.

client/src/components/Graph/Plot.vue
aschnapp: This is needed. It doesn't update the plot when things change if I use store.

PR #87, opened 2021-03-18
US303 Ascii Download Relocation

Tengelma:
### Overview
This PR contains changes relevant to Ascii data exports.

The big stuff:
- Introduces a new endpoint 'get_object_data' that takes a list of parameters and returns a JSON containing the vectors associated with those parameters for all objects in the session
- Introduces downloadData function on the client, which uses several calls to above endpoint in order to build a zip archive of TSV files in the format used by the current site

The little stuff:
- Some renaming to make things a little more explicit in what they are / what they do
- Moved download functions out of Download.vue and into a new file in utils, 'downloads.js'
- Deleted all stuff related to ASCII & keymap on server
- Tests
- Reverted astropy to 4.1 because pip could not find 4.2 when I tried to install it. Functionality seems identical for our purposes so hopefully this is ok

### Includes:
See files changed. Most of the important ones have been mentioned already.

### Developer Checklist:

### Notes:
I opted against sending labels to server and building the archive there because
- the server doesn't need to be aware of display details
- we'd basically need to send the entirety of PLOT_AXIS_OPTIONS and PLOT_AXIS_METADATA for it to work, and at that point we may as well just leave the KEYMAP in
- this gives us A LOT more freedom to customize what data the user can request / receive.


client/yarn.lock
```diff
@@ -6166,6 +6166,11 @@ ignore@^5.1.1:
   resolved "https://registry.npmjs.org/ignore/-/ignore-5.1.8.tgz"
```
aschnapp, line 6166: We should add this file to gitignore

server/requirements.txt
```diff
@@ -1,5 +1,5 @@
 astroid==2.4.2
-astropy==4.2
+astropy==4.1
```
aschnapp, line 2: why?

aschnapp (changes requested):
why change the dependency?

client/yarn.lock
aschnapp: This is just a comment. Won't reject the pr based on this.

server/requirements.txt
Tengelma: image

Mentioned this in the PR. Not sure what the solution is.

aschnapp: Ok. We can just use the old one

aschnapp (approved):

aneuhold (approved):
Looks good

aschnapp:
@aneuhold
Before you guys merge, are guys having issues with astropy 4.1? 
I cannot install it with pip over here.

aschnapp:
@Tengelma don't merge yet.

aneuhold:
> @aneuhold
> Before you guys merge, are guys having issues with astropy 4.1?
> I cannot install it with pip over here.

I did not have an issue installing astropy 4.1

aneuhold:
This is probably good to merge unless @aschnapp is still having issues. And after it has dev merged into it first.

PR #85, opened 2021-03-15
Hm306 constants ci

aschnapp:
### Overview
This moves all constant generation into CI
It does not change backend constants as the generated one still needs to be integrated. This pr does add a comment on how to use the auto generated backend constants.

### Includes:
Please see files for changes

### Developer Checklist:
- [x] The code follows the Quality Policy file on Google Drive
- [x] My code has been developer-tested and includes unit tests
- [x] I have considered proper use of exceptions
- [x] I have eliminated IDE warnings

### Notes:
This PR does not change anything to deployment. When deploying the client (whoever will work on that) you will need to generate constants before it can run correctly. 

### Refs:
- [HM-306](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-306)

PR #88, opened 2021-03-19
US280 Complex Generic Form

aneuhold:
## Overview
This PR implements User Story 280, complex generic forms. It builds on the `GenericForms` branch that Trent made. 

### Two sources of state instead of three

Previously, we had 3 sources of state or data in the application: `backend_constants`, `initial_state`, and the flattened model which was sent to the server. `initial_state` was removed in favor of reducing the places that data was held, and removed a location of duplicated code. All of the initial values were held in `initial_state` previously. 

Now, there are two primary concepts for state in the application: `HMModel`, which is equivalent to the `backend_constants`, and `HMModelFlat` which is equivalent to the flattened model which can be sent to the server. These ideas are reflected through the code. 

When the application first starts, it takes what is in `backend_constants`, and flattens it into something that can be sent to the server immediately, namely the `HMModelFlat`. It then uses `HMModelFlat` as the source state for all of the forms. This makes the code more dynamic and able to accept changes to the `hmf` library and libraries that `hmf` depends on, without changes to the front-end code. 

### `forms.js`

`forms.js` holds ALL of the configurable functionality for forms in the application now. All of the functionality that existed in the previous forms is accounted for, even the sliders. A lot of documentation exists in that file to explain what is held there. One thing to point out, is that adding new forms is as simple as copy and pasting one of the existing form objects, and editing the new form to match what is needed. The code in the rest of the application will take care of everything else. It is really that easy. If more configuration needs to be done on specific fields, that is where `parameter_properties` comes in.

### `parameter_properties.js`

`parameter_properties` was put into its own file to indicate how certain fields should be rendered, and their min/max values. This was put into it's own file separate from `forms.js` because quite a few fields are used in more than one form. That file has a lot of documentation as well in how it works. 

### Key things to look at before reviewing

Make sure to delete your current models, or `redis-cli FLUSHDB` first because of the changes in state. It also seems like a good idea to emphasize the documentation in `forms.js` and `parameter_properties.js`. Those are important to understanding how this all works. 

### Future considerations

On both dev and this branch, changes to one form causes all of the other forms to re-render. This has a bit of a performance impact on editing things. As Trent had mentioned towards the beginning we might want to consider Vuex, but now might be a better time than ever to do that. If we can implement something like that simply (emphasis on simply), then it might be good to throw in for large performance improvements. This would be a different US though and this PR would help us get very close to being able to implement that easily. 

## Includes:
- .github/workflows/client.yaml 
- client/.eslintrc.js 
- client/jest.config.js 
- client/package-lock.json 
- client/package.json 
- client/src/components/Forms/GenericForm.vue 
- client/src/components/forms/index.vue → client/src/components/Forms/index.vue 
- client/src/components/InputSlider.vue 
- client/src/components/forms/BiasForm.vue 
- client/src/components/forms/ConcentrationForm.vue 
- client/src/components/forms/CosmologyForm.vue 
- client/src/components/forms/FilterForm.vue 
- client/src/components/forms/GrowthForm.vue 
- client/src/components/forms/HMFForm.vue 
- client/src/components/forms/HODForm.vue 
- client/src/components/forms/HaloExclusionForm.vue 
- client/src/components/forms/HaloModelForm.vue 
- client/src/components/forms/MassDefinitionForm.vue 
- client/src/components/forms/ProfileForm.vue 
- client/src/components/forms/TransferForm.vue 
- client/src/constants/forms.js 
- client/src/constants/initial_state.json 
- client/src/constants/model_choices.js 
- client/src/constants/parameter_properties.js 
- client/src/main.js 
- client/src/utils/Store.js 
- client/src/utils/validators.js 
- client/src/views/Forms.vue 
- client/tests/unit/BiasForm.spec.js 
- client/tests/unit/ConcentrationForm.spec.js 
- client/tests/unit/CosmologyForm.spec.js 
- client/tests/unit/DoubleField.spec.js 
- client/tests/unit/Forms.spec.js 
- client/tests/unit/GenericForm.spec.js 
- client/tests/unit/GrowthForm.spec.js 
- client/tests/unit/HMFForm.spec.js 
- client/tests/unit/HODForm.spec.js 
- client/tests/unit/HaloExclusionForm.spec.js 
- client/tests/unit/Integration.spec.js 
- client/tests/unit/MassDefinitionForm.spec.js 
- client/tests/unit/ProfileForm.spec.js 
- client/tests/unit/validator.spec.js 
- client/yarn.lock 

## Developer Checklist:

## Notes:
- `DoubleField` was modified so that it only returns a new value when a new value is correct. This prevents unnecessary re-renders on all forms and also cleaned up the complexity a bit. This will be modified in the future when errors are propagated up to show that a form has an error inside of it. See [US87](https://the-halo-mod.atlassian.net/browse/HM-87).
- The `@/components/forms` file was renamed to `@/components/Forms` to fit with the naming scheme of the rest of the component folders. 
- `InputSlider` was added to render slider inputs.
- Unused forms and tests were removed.
- When the actual `backend_constants` file is used, another field will need to be added to the `forms.js` file to indicate the params name in `HMModel` and a different one for the `HMModelFlat`, because those are different in some cases. At the moment, they were just renamed to be the same for simplicity. 
- An option is added so that you can have a local `.env` file for the client folder that will determine if sentry should be sending out errors or not. This makes it so we can have our local dev machines sending errors locally and we don't need to keep commenting out that section of code. 
- Before sending a model to the server, it no longer needs to be flattened, but it does need to be cleaned of any extra properties that the server will not accept that come from `backend_constants`. This is defined in `cleanUpModel` in `Store.js`.
- An error in the `validators.js` file was fixed.

## Refs:
- [US 280](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-280)


aschnapp:
@aneuhold I’ll review this later today.
Does this effect the generated backend constants? #85

aneuhold:
> @aneuhold I’ll review this later today.
> Does this effect the generated backend constants? #85

It does not. But it will use them when it is merged.

client/src/components/Forms/GenericForm.vue
```diff
@@ -0,0 +1,292 @@
...
+      const doubleProps = {
+        param: parameterKey,
+        range: false,
+      };
+
+      // If properties do not exist, pass the default values
+      if (!parameterProps) return doubleProps;
+
+      if (parameterProps.plainName) {
+        doubleProps.param = parameterProps.plainName;
```
aschnapp, line 226: is there any reason why this can't this be a simple spread operator? ```doubleProps = { ...parameterProps, ...doubleProps }```

aschnapp: it would be nicer if doubleProps.param could be renamed to plainName or vice versa.

client/src/components/Forms/GenericForm.vue
```diff
@@ -0,0 +1,292 @@
...
+      const parameterProps = PARAMETER_PROPS[parameterKey];
+      if (!parameterProps || !parameterProps.options) {
+        console.error(`Error in parameter "${parameterKey}" \n`
+        + 'No options were provided for the string value parameter in '
+        + 'parameter_properties.js');
+        return {};
+      }
+      return parameterProps.options;
+    },
+    isVisible(parameterKey) {
```
aschnapp, line 267: ```return PARAMETER_PROPS[parameterKey] && PARAMETER_PROPS[parameterKey].visible != false```

client/src/components/Forms/GenericForm.vue
```diff
@@ -0,0 +1,292 @@
...
+      if (PARAMETER_PROPS[parameterKey]
+        && PARAMETER_PROPS[parameterKey].visible === false
+      ) {
+        return false;
+      }
+      return true;
+    },
+    isSlider(parameterKey) {
+      const parameterProps = PARAMETER_PROPS[parameterKey];
+      if (!parameterProps || !parameterProps.rangeSlider) {
```
aschnapp, line 277: ```return !(!parameterProps || !parameterProps.rangeSlider)```

aschnapp: isnt (!parameterProps || !parameterProps.rangeSlider) === !(parameterProps && parameterProps.rangeSlider)
this could be:
```js
return !(parameterProps && parameterProps.rangeSlider);

File: client/src/components/Forms/GenericForm.vue [hunk 4]
```diff
@@ -0,0 +1,292 @@
...
+        return false;
+      }
+      return true;
+    },
+    isSliderMin(parameterKey) {
+      const parameterProps = PARAMETER_PROPS[parameterKey];
+      if (!parameterProps || !parameterProps.rangeSlider
```
aschnapp, line 284: this could call the previous function

aschnapp: in the series above you're using boolean operators to then return a boolean, which is what the operators get evaluated to. You can just return what (or !) the operators evaluates to.

client/src/main.js
```diff
@@ -8,14 +8,16 @@ import Store from './utils/Store';
...
  */
-Sentry.init({
-  Vue,
-  dsn: 'https://7a99f3aaaa144241960d3857a1f4dee9@o516709.ingest.sentry.io/5623500',
-  integrations: [new Integrations.BrowserTracing()],
-  tracingOptions: {
-    trackComponents: true,
-  },
-});
+if (process.env.VUE_APP_SENTRY_ON !== 'FALSE') {
```
aschnapp, line 11: 👍

.github/workflows/client.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       with:
         # npm cache files are stored in `~/.npm` on Linux/macOS
         path: ~/.npm
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+        key: ${{ runner.os }}-build-v2-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
```
aschnapp, line 35: why v2?

aschnapp (changes requested):
The forms that's autogenerated have some parts that looks a little odd (see image below)
it would be nice if there is some rearrangement to that. We can add that to a different user story. I dont think that's part of this US. 
Please look at the other comments for the requested changes
Screen Shot 2021-03-21 at 2 43 35 PM

.github/workflows/client.yaml
aneuhold: This was a change that @Tengelma made. It was an issue with the cache and it is the only way to reset the cache evidently for Github actions. It is a known problem with github actions.

client/src/components/Forms/GenericForm.vue
aneuhold: Done! Great idea. That simplified things.

client/src/components/Forms/GenericForm.vue
aneuhold: This was implemented, with a slight alteration to match the correct logic.

client/src/components/Forms/GenericForm.vue
aneuhold: Implemented with a slight variation to match correct logic. 😄

client/src/components/Forms/GenericForm.vue
aneuhold: Implemented 😄 Great ideas! Much cleaner now.

aneuhold:
@aschnapp trying to think of another way to format it and having a bit of a hard time. Do you have any ideas? It could be another US, but if it could be implemented here then I will try and do it 🙂 .

vperuzzi (approved):
Looks great, Anton. The consolidation of the sources of state is a significant improvement. I really haven't worked on the forms much, so I don't really have any specific feedback or ways to improve what you wrote. As you noted, I think your changes are very conducive to adding Vuex in the next sprint, so this is something we should discuss. Again, great work. Approving.

client/src/components/Forms/GenericForm.vue
```diff
@@ -0,0 +1,268 @@
...
+      const parameterProps = PARAMETER_PROPS[parameterKey];
+      const doubleProps = {
+        plainName: parameterKey,
+        range: false,
+      };
+
+      // If properties do not exist, pass the default values
+      if (!parameterProps) return doubleProps;
+
+      return { ...Object.assign(doubleProps, parameterProps) };
```
aschnapp, line 225: The spread here is redundant.

aschnapp (approved):

aschnapp:
@aneuhold approved, with just one minor change. 
Your new changes looks good.

client/src/components/Forms/GenericForm.vue
aneuhold: Thank you 😄

PR #96, opened 2021-03-22
US302-Task 309: Import External SVG CSS

vperuzzi:
### Overview
This PR implements Task 309 of US 302 Plot Image Functionality, adding all relevant CSS external styles for the `plot-svg` element inline. The returned SVG from the download plot image is now styled the same as the svg rendered in the browser. Note: I was unable to successfully complete the other task in this user story, [Task 308 - Export SVG to PDF](https://the-halo-mod.atlassian.net/browse/HM-308) due to conflicts with MathJax. I elaborate on these findings in the **Notes** section.

SVG before:
image

SVG after inline styles added:
image

### Includes:
- client/src/utils/inlineCSS.js
- client/src/Components/Download.vue

### Developer Checklist:

### Notes:
As noted earlier, I was unable to successfully export the SVG to PDF. While there exists numerous methods of exporting an SVG to a PDF, namely through the process of: SVG --> Canvas --> PDF, conflicts with MathJax made every attempt futile. I tried several different libraries including svg2pdf, Canvg, JSpdf, pdfkit, and they all resulted in errors. After doing some digging, I came across a few relevant Stackoverflow threads that shed light on why this was occurring. See:
- https://stackoverflow.com/questions/42785850/render-mathjax-into-pdf
- https://the-halo-mod.atlassian.net/browse/HM-308?focusedCommentId=10011
We should discuss with Steven what other viable alternatives there are. I think the returned SVG is suitable for most usecases, but at this time, it seems to be nearly impossible to embed the SVG with MathJax into a PDF.

### Refs:
- https://the-halo-mod.atlassian.net/browse/HM-309


aneuhold (approved):
Looks great and works great!

Only issue might be the `/* eslint-disable */` part. But that is probably for a good reason. Sometimes manual editing of dom things like that aren't up to par with ESlint and there isn't always a solution.

aneuhold:
@vperuzzi it looks like since the merge this now downloads the old plot SVG.

aneuhold (changes requested):
It looks like a lot of existing files were deleted for some reason. Was this intended?

vperuzzi:
Closing, as all remaining conflicts have been resolved in #151

PR #144, opened 2021-03-28
US321 Complete Deployment

aneuhold:
## Overview
This PR implements US 321, the complete deployment of the project.

### Client deployment
The site is successfully deployed to Netlify, and all pull requests should have an added comment by the Github action that does that. For example, there should be a comment below with the front-end deployed for this branch ⬇️. The tactic that we discussed at the meeting of using Github Actions to do all the build steps and sending the completed bundle to Netlify worked. So that was done which means we have 1 second builds on Netlify, so we will likely never run out of minutes there. This also means that you can deploy the client for every single commit if you feel so inclined. 

### Server deployment
https://api.thehalomod.app works now. It was an HTTPS issue with Flask. Also, the front-ends deployed on Netlify work with the backend automatically. You can use the live server as the backend to your local front-end by setting the following in a `.env` file in the `client` folder:

```
VUE_APP_SERVER_URL=https://api.thehalomod.app
```

## Includes:
- .github/workflows/client.yaml 
- .github/workflows/clientDeployment.yaml 
- .github/workflows/server.yaml 
- .github/workflows/serverDeployment.yaml 
- README.md 
- client/src/env.js 
- generate_constants.sh 
- server/Dockerfile 
- server/README.md 
- server/config.py 
- server/docker-compose.yml 
- `server/halomod_app/__init__.py`
- server/requirements.txt 
- server/run.sh 

## Developer Checklist:
- N/A - My code has been developer-tested and includes unit tests - It has been tested extensively manually but there are not specific unit tests. 

## Notes:
- The READMEs were updated with a bit more info. 
- There was a pretty big issue with the session cookie not being set when using the live server backend. This has been fixed with a workaround, and documented in the code. It is an active issue with Flask-Session. There is even a PR in for it, but they haven't merged it yet for some reason.
- The `stop_signal` was modified for the docker container so that when using `docker stop <container-ID>` it stops Flask gracefully. 

## Refs:
- [US 321](https://the-halo-mod.atlassian.net/jira/software/projects/HM/boards/1?selectedIssue=HM-321)


aneuhold:
Gotta work on fixing the constants generation

aneuhold:
Okay, things seem to be fixed now.

README.md
```diff
@@ -34,3 +34,5 @@ The only parts that are not included in the server and client at this moment are
 - Run tests on both the server and client when new pushes are made to any branch in the project
 - Run linting on both the server and client when new pushes are made to any branch in the project
 - Deploy the client when a new push is made to the `main` branch 
+
+Note that in order to execute shell files (files with an `.sh` extension), then need to be ran in Github Actions with `sh shell_file_name.sh`. If they are ran with `. ./shell_file_name.sh` they will not run correctly.
```
Tengelma, line 38: This is a little ambiguous. It seems to imply that I can't run the shell script unless I'm doing so through Github Actions (which shouldn't be true). Could you rephrase it a bit?

server/README.md
```diff
@@ -24,25 +24,40 @@ Before doing any of the commands below, make sure to install the Python packages
 1. In the first shell window, run `. ./run.sh --dev`
 1. The server should now be located at `localhost:5000`
 
+If an issue comes up where it says that the port is already used, check to see if redis is running in another terminal somewhere. Or it might have been installed as a service and running in the background, which means it does not need to be started again.
+
+Sometimes while developing it can be useful to clear the redis DB of data if issues arise with the models. This can be done in a separate terminal than that which is running redis by typing `redis-cli FLUSHDB`.
```
Tengelma, line 29: Redis is usually run as a service so you don't need a terminal open. I know this is mentioned here but I think the language should treat that as the default thing to look at it. After all its pretty obvious if you have a terminal open that's running redis and that's whats causing the conflict. It's less likely that they'll pick up on a background service they might've had running for the last year unknowingly.

It's a little change but something I'd encourage.

server/halomod_app/__init__.py
```diff
@@ -42,15 +43,41 @@ def create_app(test_config=None):
...
     app.config.from_object('config.Config')
 
-    CORS(app, origins="http://localhost:*", supports_credentials=True)  # enable CORS
+    # The different origins that the server will allow connections from.
+    # These are specified as RegEx.
+    origins = [
+        re.compile(r'http\:\/\/localhost\:.*'),
+        re.compile(r'https:\/\/.*thehalomod\.netlify\.app.*'),
+        re.compile(r'https:\/\/.*thehalomod\.app.*'),
+    ]
```
Tengelma, line 52: :fire:

Tengelma (approved):
Obviously works (because I can go the relevant sites) so green light there. Cool to have this up for sure.

I did have a couple little comments on some of the documentation you added and would like to see those changes made but you have my approval.

README.md
aneuhold: You are right that is vague. Just updated it to the below:

Note that in order to execute `.sh` files inside of a workflow (the `.yaml` files in `.github/workflows`), they need to be ran with `sh shell_file_name.sh` INSTEAD of something like `. ./shell_file_name.sh`. If they are ran with `. ./shell_file_name.sh` they will not apply to the environment correctly. 

Does that make more sense or should it be a little more descriptive or changed? Trying to think of the right wording to make it better so suggestions are welcome.

Tengelma: That makes more sense for sure! Thanks for the fix.

server/README.md
aneuhold: Changed it to the below:

If an issue comes up where it says that the port is already used, check to see if redis is running in another terminal somewhere, or in a background service. By default redis runs as a service, so it might not be necessary to run it in another terminal if it is.

Does that sound good or what should it be?

Can't remember how I installed it but it doesn't run as a service on my computer and it is started manually. Didn't constantly want it running in the background so I might have chosen to do that specifically.

Tengelma: What about:

If an issue comes up where the port used by redis is already in use, you may already have it running through a background service or an open terminal. You can check to see if redis is already running with 'X' command. If it is, you're good to go and don't need to run it again. If it's not, you may have some other service or application active that may be making use of this port. If this is the case, here's a link to instructions on how to configure redis to use a different port: <link>.

Sometimes while developing it can be useful to clear the redis DB of data if issues arise with existing models. This can be done in any inactive terminal with the command 'redis-cli FLUSHDB'.

PR #146, opened 2021-03-31
Hm87 form errors

aschnapp:
### Overview
This currently verifies double field errors, but does not verify any null values for dropdowns or invalid sliders as they're assumed to be correct.

### Includes:
- see files tab

### Developer Checklist:

### Notes:
Form generation needs to be fixed, it's generating null values for dropdowns.

### Refs:
- [HM87](https://the-halo-mod.atlassian.net/browse/HM-87)


aneuhold (changes requested):
Right now it looks like there is a bunch of console logging that is coming from somewhere:

image

image

Once those are removed then it looks like it works well!

aschnapp:
@aneuhold Ah forgot to remove them. Done!

aneuhold:
@aschnapp were they pushed? It doesn't look like any other commits have been pushed to the branch. Haha

aschnapp:
@aneuhold Sorry, forgot type my ssh password :( 
They should be now.

aneuhold (approved):

Tengelma (changes requested):
So this doesn't quite work atm. Parameters associated with model selection don't throw validation errors and trigger the highlighting the same way other parameters do. Note the faulty growth subform data in the image below.

![image](https://user-images.githubusercontent.com/25331712/113496860-b2d00980-94b2-11eb-9aaa-5a1fc2debe63.png

This definitely needs to be resolved!

I also think deactivating the button when in an invalid state isn't very informative in terms of what actually needs to be done. I think it would be better to have it either:
- scroll to the top and display an error there that says what forms have issues
- display a modal that indicates the error and says what forms have issues

These should be pretty simple fixes. Looks good otherwise!

aschnapp:
@Tengelma Added a tooltip, and uploaded a video. I'm unable to reproduce the errors that you're reporting.

aneuhold (changes requested):
Just double checked what @Tengelma had mentioned and it looks to be similar but a slightly different issue on this end. Currently when a model parameter is wrong, but then the model is changed to something else and the new model parameters are correct, it still shows that the form has an error. See below:

image

It seems like this might have been intended behavior, but if the user switches to a different model and enters the correct fields there, then it will probably be confusing for them to have to find out which previous model they had messed up on to go back to. Also, because the server is only sent the parameters for the currently selected model, the error in the previous one won't actually cause any issues when the data is sent to the server, so it is a bit misleading to still mark it as incorrect. 

Could this be adjusted?

aschnapp:
@aneuhold It's been fixed

aneuhold (approved):
Looks good!

Tengelma (approved):
I'd still like something more descriptive than 'fix errors in red' because there's no reason not to. I know I'm being nitpicky on this but may as well imitate every form in existence instead of reinventing the wheel with how this is indicated.

Regardless, I couldn't replicate the issue I was having before so just going to assume it works 99.9% of the time and approve.

PR #147, opened 2021-04-01
US322 Form Performance Optimization [PARTIAL]

Tengelma:
### Overview
This is a first pass at performance optimization and focuses on GenericForm specifically. My goal going into it was to front-load the work by reformatting and splitting backend_contstants programatically so that the logic required to make forms work was less intensive. The main changes are as follows:
- backend_constants.js, which previously just exported an identical copy to the backend_constants generated on deployment, now splits that content into DEFAULT_FORM_STATE and FORM_OPTION_DEFAULTS, which are well formatted representations of our initial form state & all the model-option-specific defaults respectively. Both of these now tie the relevant information to the id of the form it corresponds to so that it can be referenced / used without knowledge of how the generated constants look
- as a result of the above, I was able to remove all of the javascript logic from forms.js, with the only added information being a 'coreParams' key that contains an array of parameter names that don't belong in the <form>_params object and aren't the <form>_model. Some of these functions were being called on every render and now aren't required at all.
- The GenericForms component now uses the formId passed in from the component above it to 'look itself up' in forms.js and save that metadata locally on creation, preventing the need to rebuild props on renders. It also handles the few lines of special logic 'cosmo' requires directly, as that it the only exception to the way the rest of the forms work. variable names have also been simplified a bit to reflect the simpler format of state.
- Store has been updated to reflect changes.
- Tests have been updated to reflect changes.

### Includes:
See files changed.

### Developer Checklist:

### Notes:
Big concern is that despite these changes, it doesn't really feel much snappier because the drawer on forms is the thing sucking up all the resources. Here is a performance profile with the drawer included:
image
and here is a performance profile without it, where I got the same amount of time spent on the 'FormWrapper' component:
image

Note that the md-list-items are the most resource intensive item and take up almost twice as much time as our forms. This is because they're also router-links and have a ripple animation associated with them that consumes resources on every update to form state for some reason. I will be digging into this for the remainder of the sprint to see if I can prevent this rerender, potentially disable the ripple effect, or maybe even find a drawer component from another library that is more efficient.


client/src/constants/forms.js
```diff
@@ -79,819 +58,122 @@ debug.enabled = false;
...
- */
-
-/**
- * The primary source of information on the client side which will generate
- * the forms. This contains everything that is needed to configure the
- * front-end forms.
- *
- * @type {Forms}
- */
 const FORMS = {
```
aneuhold, line 68: The reason for specifying the type is so that autocomplete works in other parts of the application that use this object.

client/src/constants/forms.js
```diff
@@ -40,30 +24,25 @@ debug.enabled = false;
...
- * retrive properties for the different model selections. This is sometimes
- * different than what is sent to the server.
- *
  * @typedef FormDetails
  * @type {{
  *  id: string,
  *  title: string,
- *  model_key: string | undefined,
+ *  modelKey: string | undefined,
+ *  coreParams: String[]
```
aneuhold, line 43: `coreParams` looks like it can be a `string[]` or `undefined`.

client/src/constants/forms.js
```diff
@@ -1,27 +1,11 @@
+/* eslint-disable */
```
aneuhold, line 1: Does ESLint need to be disabled for the whole file?

client/src/utils/Store.js
```diff
@@ -99,6 +63,18 @@ export default class Store {
    * }}
    */
 
+  /**
+   *
+   * @param {object} model
+   */
+  flatten = (model) => {
```
aneuhold, line 70: Might be nice to have a description here.

client/src/constants/backend_constants.js
```diff
@@ -1,11 +1,89 @@
...
+  hmf: 'FittingFunction_params',
+  hod: 'HOD_params',
+  bias: 'Bias_params',
+  halo_concentration: 'CMRelation_params',
+  tracer_concentration: 'CMRelation_params',
+  halo_profile: 'Profile_params',
+  tracer_profile: 'Profile_params',
+  halo_exclusion: 'Exclusion_params',
+  transfer: 'TransferComponent_params',
+};
```
aneuhold, line 24: Could these be put in the `forms.js` file? It looks like something that is specific to each form.

client/src/constants/backend_constants.js
```diff
@@ -1,11 +1,89 @@
...
+
+const BACKEND_CONSTANTS = generatedBackendConstants;
+const form_option_defaults = {};
+const default_form_state = {};
+
+/* halo_model is weird */
+form_option_defaults.halo_model = {};
```
aneuhold, line 31: Because the backend constants file is where behavior of specific forms is determined now, it would probably be good to put a description of how this behavior needs to be accounted for, and/or to store it in `forms.js` because it looks like this is specific to each form. As it stands, the configuration for each form is being spread among two files. There are a couple ways to account for that. Either with comments, or storing it in one place. But either way, a further description will be needed on how to do this with other forms in the future. Same goes for cosmo in this same file.

client/src/constants/backend_constants.js
```diff
@@ -1,11 +1,89 @@
-import deepFreeze from 'deep-freeze';
 import generatedBackendConstants from '@/../generated/backend_constants.json';
+import forms from '@/constants/forms.js';
+import deepFreeze from 'deep-freeze';
 
 /**
  * Note that these are generated by running `npm run setup` first.
  *
  * @typedef HMModel
```
aneuhold, line 8: if HMModel is no longer being referenced anywhere in the application, this type can likely be removed.

aneuhold (changes requested):

client/src/constants/backend_constants.js
Tengelma: I was actually able to remove the halo_model mention which is nice. Only exception now is cosmo, which I've documented.

I am a little confused by this feedback, though? There's not really any 'behavior' in backend_constants. It really just reformats the constants generated in the startup script to something that corresponds to client side representation of state instead of server side representation of state. That's all.

Also, I'm not sure what you mean by configuration information being spread across multiple files? Backend constants is exporting the exact same information as before (cosmo defaults exception was included in here before, too), it's just in a slightly different format. Can you clarify?

client/src/constants/forms.js
Tengelma: Was a straggler. Fixed.

client/src/constants/forms.js
Tengelma: :heavy_check_mark:

client/src/constants/forms.js
```diff
@@ -79,819 +58,122 @@ debug.enabled = false;
...
- *
- * @typedef GetModelChoicesDataFromFlatFunction
- * @type {(hmModelFlat: HMModelFlat) => ModelChoicesData}
- */
+ /**
+  * Defines the parameters associated with the subform common to all model
+  * selections associated with that subform.
+  * @typedef CoreParams
+  * @type {Array.String}
+  */
```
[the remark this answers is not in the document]
Tengelma: Removed this. Originally planned to have it as a custom type then realized it was too basic to warrant it.

client/src/constants/forms.js
Tengelma: Added it. Was an accidental deletion.

client/src/utils/Store.js
Tengelma: Done.

client/src/constants/backend_constants.js
Tengelma: Removed.

Tengelma (commented):
Fixed almost everything. Had questions on a couple of items, though.

client/src/constants/backend_constants.js
aneuhold: No that does make sense. Maybe just a short description of what it is for above it so that if it needs to be changed if backend constants from the server changes then it will be easy for Steven to understand what it does.

client/src/constants/backend_constants.js
aneuhold: You had explained the reasoning in the other comment. This makes sense.

aneuhold:
Just one thing left, which is easy because it is just commenting. It does seem more performant for sure, and a lot cleaner.

client/src/constants/backend_constants.js
Tengelma: Updated to provide a description.

aneuhold (approved):

dblevin1 (approved):
Works Great!

PR #151, opened 2021-04-03
Svg Inline Styles Fix

vperuzzi:
### Overview
This PR encapsulates changes that were introduced in #96, but resolves all existing merge conflicts that were present when trying to merge dev into that branch. AddInlineCSS method was added to `downloads.js`.

### Includes:
- client/src/utils/downloads.js

### Developer Checklist:

### Notes:
N/A

### Refs:
- https://the-halo-mod.atlassian.net/browse/HM-309

### DIFF
--- a/client/src/utils/downloads.js
+++ b/client/src/utils/downloads.js
@@ -37,24 +37,23 @@ export async function downloadData() {
   for (const [kind, options] of Object.entries(PLOT_AXIS_OPTIONS)) {
     /* Gets relevant fields from constants files */
     let params = [kind, ...options.y];
-    if (kind === "m") params = params.filter(value => value !== "how_big");
-    const labels = params.map(param => PLOT_AXIS_METADATA[param].label);
+    if (kind === 'm') params = params.filter((value) => value !== 'how_big');
+    const labels = params.map((param) => PLOT_AXIS_METADATA[param].label);
 
     /* API request */
     const response = await axios.post(`${baseUrl}/get_object_data`, {
-      param_names: params
+      param_names: params,
     });
     const json = response.data;
 
     /* Construct CSV */
     Object.entries(json).forEach(([name, parameters]) => {
       const data = unzip(Object.values(parameters));
-      const csv = `${labels.join("\t")}\n${data.map(row => `${row.join("\t")}\n`).join()}`;
+      const csv = `${labels.join('\t')}\n${data.map((row) => `${row.join('\t')}\n`).join()}`;
       zip.file(`${name}_${kind}_vector.csv`, csv);
     });
   }
   /* eslint-enable */
-
   const blob = await zip.generateAsync({ type: 'blob' });
   return window.URL.createObjectURL(blob);
 }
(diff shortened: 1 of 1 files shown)


aneuhold (approved):
Looks great! Nice and clean 😄

aneuhold:
Should #96 be closed now?

client/src/utils/downloads.js
```diff
@@ -12,19 +37,19 @@ export async function downloadData() {
...
-    
+    if (kind === "m") params = params.filter(value => value !== "how_big");
+    const labels = params.map(param => PLOT_AXIS_METADATA[param].label);
+
     /* API request */
     const response = await axios.post(`${baseUrl}/get_object_data`, {
-      param_names: params,
+      param_names: params
     });
     const json = response.data;
```
Tengelma, line 47: Some of these style changes seem wrong to me? I think the linter / style guidelines call for single quoted strings, parentheses for function shorthands, trailing commas on objects, etc so not sure what happened here. eslint was disabled for this block so I could use the for...of syntax (forEach wouldn't work) but the rest should be reverted to the usual rules :smile:

Tengelma (changes requested):
Looks good. Little style mishap on the style PR though haha

Once that little thing is fixed I'll approve!

client/src/utils/downloads.js
aneuhold: Just ran my own linter on it and it looks good from this end. Is it a specific rule on your side @Tengelma? It should be following what is in `.eslintrc` for all of us.

Tengelma: The issue is the /* eslint-disable */ at the top of this block of code -- its preventing the linter from catching the little issues so even though it seems to pass there are still some stuff that wasn't intentionally allowed. I was the one that put the thing in there because:
- Current style guidelines don't like for...of but it was 100% necessary here
- Current style guidelines don't like awaits in loops which was 100% necessary here
but it looks like this section got autoformatted to a different set of style guidelines because linting was disabled for the whole block. Probably just need to do more targeted exceptions to the linting rules.

aneuhold: Ah, ya, targeted ignores are always better.

vperuzzi: > 
> 
> Should #96 be closed now?

Yes, I have closed it.

vperuzzi:
> 
> 
> Should #96 be closed now?

Yes, I have closed it.

client/src/utils/downloads.js
vperuzzi: Yeah, a lot of the lint fixes get overwritten I believe when I first create a branch off of dev. I am not sure why; I have resolved these lint issues with the latest push. Please re-review at your earliest convenience :)

aneuhold: It looks like the exact things mentioned are fixed. So it seems like a good idea to just resolve this one and pull it in now before the meeting. Doing that now :)

PR #155, opened 2021-04-07
U339 Reject Same-Name Models

Tengelma:
### Overview
This implements a 'name in use' validation check for names in the model list & in the create dialog. The check in the model list was a bit quirky and I had to create my own error component instead of leaning on md-field's to avoid issues with the error being hidden by other components (yes, I did tinker with z-values and a number of other solutions before deciding to go this route). This meant making each item in the list a double item so that it could have subtext.

### Includes:
Check files changed.

### Developer Checklist:

### Notes:
- Might be worth introducing some other validation checks (no special characters, no empty field, etc...) but could let Steven do this to get him familiar

### DIFF
--- a/client/.eslintrc.js
+++ b/client/.eslintrc.js
@@ -29,6 +29,7 @@ module.exports = {
      * have a leading underscore.
      */
     'no-underscore-dangle': [0],
+    'no-irregular-whitespace': [0],
   },
   overrides: [
     {
--- a/client/src/components/Models/Model.vue
+++ b/client/src/components/Models/Model.vue
@@ -15,7 +15,7 @@
       <span id="error">{{ invalidName ? 'Name is already in use' : '⠀'}}</span>
     </div>
 
-    <div v-if="editing">
+    <div v-if="editing" id="model-buttons">
       <md-button
         class="md-icon-button"
         @click="reset">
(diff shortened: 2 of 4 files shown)


vperuzzi (approved):

aneuhold (changes requested):
Everything seems to work well but the model name seems to no longer be centered:

image

aneuhold (changes requested):
It looks like it is aligned with the buttons now, but it isn't centered in the list item:

image

Tengelma:
> It looks like it is aligned with the buttons now, but it isn't centered in the list item:
> 
> image

This is intentional. There is a placeholder span beneath it so that it doesn't reformat when the error message pops up. The list item contains space for both.

aneuhold:
@Tengelma could the placeholder span be set to `display:none;` when there isn't an error?

aneuhold:
Its just that it is something that people see first when they open the website up, and having something off-center like that gives it a bad look right off the bat.

aneuhold:
@Tengelma maybe we could have something like this that comes up when the name is already in use? This would allow it to stay centered:

image

Code on this page: https://vuematerial.io/components/input/

aneuhold (approved):
Alright

Tengelma:
> Alright

Btw, the reason I wasn't seeing what you were pointing out was my monitor. It looks flat white on one monitor and then a distinguishable white box on the other. Kinda weird :man_shrugging:

aneuhold:
@Tengelma oh okay. That makes sense then why it would look fine if it had another box around it.

Issue #139, opened 2021-03-27
Initial Model


steven-murray:
Currently when starting up, there is no model displayed. I think it's great to just have a pre-computed default model displaying when you go to the page. It also makes the next computation slightly faster, since a bunch of stuff is already computed.

aneuhold:
Just for future consideration, this would involve changing how the client-side store works in that it needs to send over what the client thinks it has and compare that with what the server has to reconcile the differences in the models. This should be done anyway, because if the server has something different than the client right now, it just doesn't show a plot but shows that there are models there. It should reinstantiate those models with the server if they don't exist yet on the server.

This previous step is required because the server will need to know if the client has any models yet to provide a default one. It wouldn't be good for the server to assume that because it doesn't have any models stored for that session ID, that it should provide a new one because it could overwrite the user's model or confuse the user because an extra model is sent back to add on top of their list.

aneuhold:
Decision is to have one model on the server pre-computed and just clone that.

PR #161, opened 2021-04-19
Add default tracer halo model

vperuzzi:
### Overview
This PR implements a change to include a precomputed default Tracer Halo Model on the server, as detailed in #139 that is passed as the `hmf-driver`'s `previous` parameter in the create `/model` endpoint. The default model is based on the following parameters:
```python
TracerHaloModel(rmax=150, rnum=200, transfer_params={"kmax":1e3, 'extrapolate_with_eh': True})
```

### Includes:
- server\halomod_app\endpoint_model.py

### Developer Checklist:

### Notes:
Steven has articulated that any new models that are cloned from this initial model that alter the cosmology parameters shouldn't use the same parameters as the default model, as the calculation would take too long. This additional constraint/consideration is outside the scope of this one-point user story, so it is *not*  implemented. A new GitHub issue should be created to address this in the future. 

### Refs:
- https://the-halo-mod.atlassian.net/browse/HM-340

### DIFF
--- a/server/halomod_app/endpoint_model.py
+++ b/server/halomod_app/endpoint_model.py
@@ -7,6 +7,7 @@ from flask_cors import CORS
 from flask import Flask, jsonify, request, session, abort, send_file
 from . import utils
 from .utils import get_model_names
+from halomod import TracerHaloModel
 import base64
 import json
 import dill as pickle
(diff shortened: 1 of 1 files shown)


aneuhold (approved):
Simple and looks good! Nice :)

aschnapp (approved):

PR #164, opened 2021-04-19
Icons

steven-murray:
### Overview
Adds some new SVG icon options. Would like some input as to which looks best...

### Includes:
- client/src/assets/thm_icon_x.svg
-
### Developer Checklist:


aneuhold:
The icons might not need to be changed to reduce their size. Still working on analyzing the build, but the icon may no longer be the largest thing pulled. Just felt like a good idea to say something before you spend time on it Steven 😄

aneuhold:
Here is the result when running the site throttled to slow 3G with no caching:

image

`tex-svg` takes up the most space, so the new svg should be completely fine. No downsizing needed. So now we just need to choose which one is best 😄 My own vote in order of preference is:

1. `thm_logo_0.svg`
2. `thm_logo_1.svg` - Noting that this one will require a color scheme change it seems. 
3. `thm_logo_2.svg`

aschnapp:
My order of preference is: 
1. ```thm_logo_1.svg``` - The background and text contrast is much better compared to the other ones. 
2. ```thm_logo_0.svg``` - Background seems a little too light compared to the text, making it look kind of dull. 
3. ```thm_logo_2.svg```

steven-murray:
Oh, nice work @aneuhold! 19 seconds is a bit scary for the tex.js! Makes me wonder if we can reduce that somehow.

I think I'm going to cast my vote for `thm_logo_0.svg`. I do like the contrast in `_1`, but I like the simplicity of `_0`. @aschnapp do you think there's a way to tweak `_0` to improve it?

aneuhold:
@steven-murray would you like me to go ahead and start implementing the new logo? Namely `thm_logo_0.svg`.

steven-murray:
Hi @aneuhold, yeah I think that's good. I haven't had a chance to play with it any further. If I do, I can always update later.

Issue #115, opened 2021-03-26
Breaking CAMB shuts down the whole server


steven-murray:
CAMB is a library that computes the transfer function (by default). It is written in FORTRAN, and if it really breaks, Python doesn't seem to be able to handle it as an exception -- the whole server just crashes. This is obviously not desirable. One way to mitigate this is to get the right parameter constraints in for cosmology, on the forms (these don't seem to be in yet). I'm not sure how else to protect against this... it might just be that we need a heartbeat program on the server to restart it if its down.

aneuhold:
H0 could be set to 1000 to try and crash it. Or a negative number. It should be between 50 and 100.

PR #172, opened 2021-05-09
Auto restarts server

aschnapp:
### Overview
Automatically restarts docker container if container crashes


aschnapp:
Should fix some parts of #115

aneuhold:
@aschnapp it doesn't look like this brought the server back up for some reason. Even after triggering the server deployment it is still down.

steven-murray:
@aschnapp is there some way we can bring the server back up manually for now?

PR #173, opened 2021-05-21
nicer-enter

steven-murray:
### Overview
This just adds the functionality that pressing ENTER on the main form submits the form.

### Includes:
- Enumerate the file-level changes which were introduced
- Forms.vue -- implemented three things into the form, which I found on https://forum.vuejs.org/t/capture-keypress-for-all-keys/14560/12
- path/to/file2.java

### Developer Checklist:
- [ ] I have considered proper use of exceptions


aneuhold:
@steven-murray just added some conditional logic for when to allow the enter button to trigger an action. Before it was triggering a save from the main screen so now it should only trigger when in the forms view. Also when the form is valid.

aneuhold (approved):

PR #175, opened 2021-06-04
Server Client Model Syncing

### LINKED ISSUE #174: Server-Client Syncing of Models
Currently, when the server restarts, any models currently held on the client side don't sync back up with the server. 

This should be changed so that even if the server restarts, the client can send over data that will allow the server to rebuild those models so the user doesn't lose their data and so it isn't confusing when there are models without associated plots on the screen.

aneuhold:
## Overview
This PR closes #174 so it implements a way for a client to interact with the server on startup and add models to the server if they aren't already there. This solves the issue of the server restarting and user's models no longer existing because they are added back to the server when the client visits the site again.

In the case the user doesn't have any issues with the server having something out of date, nothing different occurs. In the case that the server does need some models added, the user is presented with the following loader:

image

Once the models are added back to the server, and the plot is rendered, this loader goes away. This makes it so the page doesn't continually load and the user is left wondering what is happening in the background. 

### Initial Model Generation Update

Now when the initial model is created, it saves a file with the object contained within it. Just pickled is all. This way the server checks if the file is there on restarts and uses that instead. This saves on the refresh time of the server dramatically, especially when developing. 

### Flask update to 2.0

An update needed to be done to Flask to allow synchronous operations. It used to be 1.0, but 2.0 allowed threading tooling so it was updated. This was needed to add a [semaphore](https://en.wikipedia.org/wiki/Semaphore_(programming)) which only allows one model to be created at a time if multiple requests are coming in. This is because there is a consistent Fortran crash when two models are being generated at the exact same time. So the semaphore is placed through the endpoints to only allow one to occur at a time. 

## Includes:
- .gitignore 
- client/src/components/Graph/index.vue 
- client/src/utils/Store.js 
- client/src/utils/axiosHelper.js 
- client/src/views/Home.vue 
- server/README.md 
- server/halomod_app/endpoint_model.py 
- server/halomod_app/endpoint_models.py 
- server/halomod_app/utils.py 
- server/requirements.txt 
- server/tests/test_models_endpoint.py 

## Developer Checklist:

## Notes:
- A quick helper file was added for axios so it can be used in other files with the same settings if wanted
- An entry was added to the readme about a strange numpy error. Not sure if it will happen again or to others, so added it there just in case with a fix.
- Setup flask to go into debugging mode if the environment is set to "development"
- Updated some of the server comments so they made more sense


client/src/utils/Store.js
```diff
@@ -103,6 +109,91 @@ export default class Store {
     }
   }
 
+  /**
+   * Syncs the models on the client with the models on the server. If the
+   * client has models the server doesn't have, those are added to the server.
+   */
+  syncModels = async () => {
+    try {
+      // Get the model names from the server. Model names are used as equality
```
steven-murray, line 118: So, for a particular user, the model names must all be unique. However, different users could have the same model names -- will this break things?

server/halomod_app/endpoint_models.py
```diff
@@ -102,28 +88,84 @@ def get_models_data_toml():
...
+            **params)  # creates model from params
 
-"""Clone model
-PUT /models
+    modelCreationSem.release()
 
-Parameters:
-- model_name: string
-- new_model_name: string
+    print('it got here 2')
```
steven-murray, line 145: I assume this is meant to be removed :-)

steven-murray (approved):
Thanks @aneuhold, this is really important work! I've made only a couple of comments which I think should be addressed before merging, but pre-approving because they should be simple enough.

client/src/utils/Store.js
aneuhold: It will not because the only model names returned from the server are for the particular user session from which the request is coming from. So when hitting that endpoint you only get model names for models you have locally on that browser.

server/halomod_app/endpoint_models.py
aneuhold: Ah yes, good catch! 😂

PR #176, opened 2021-06-12
I100 Model Naming

### LINKED ISSUE #100: Where you name the model
At the moment, when you create a new model, it will only ask you to name the model after you click "Create". I intuitively thought I should be able to modify the model name in the form itself here:

image

where it says "New Model". Not sure if this is difficult architecture-wise (you'd still need the popup in case the user doesn't set the title there...). If not, I'd suggest adding it.

### LINKED ISSUE #121: It's not clear how to edit the model names
Indeed, I thought you couldn't, and was going to make an issue about it, but then accidentally clicked on one of the names in the Model box. Now that I know it's there, it makes sense, but it's not visually obvious that you can do it. 

I guess this could be fixed by #100 as well.

aneuhold:
## Overview
This PR resolves #100 and resolves #121 by implementing a few visual additions and also adding a way to rename models within the `Forms` component without hitting "Enter" first.

### Renaming models within `Forms` component

The new way to rename models within the `Forms` component looks like so:

image

which also gives the same type of error info when a model already has the name chosen, and it disables the "Save" button when it is incorrect.

image

It also does this when creating a form:

image

### Renaming models from main page

There is now a tooltip that shows up when you hover over a model name that says that the model name can be edited by clicking on it there:

<image src="https://user-images.githubusercontent.com/17774559/121787367-856e9e80-cb7a-11eb-943b-f70cc55364b0.png" width=500 />

In addition, as you can see above, the model names are centered again. One other change is when a model name is incorrect, instead of having text come up at the bottom (causing issues with the text moving around) it comes up on the side with a very visible icon:

<image src="https://user-images.githubusercontent.com/17774559/121787536-6d4b4f00-cb7b-11eb-9d92-a076fa144995.png" width=500 />

## Includes:
- .gitignore 
- client/src/components/Forms/GenericForm.vue 
- client/src/components/Forms/index.vue 
- client/src/components/Models/Model.vue 
- client/src/components/Models/index.vue 
- client/src/utils/Store.js 
- client/src/views/Forms.vue 

## Developer Checklist:
- [ ] My code has been developer-tested and includes unit tests - It has been tested manually but does not have unit tests.

## Notes:
- Updating how `DoubleField` handles validation solved a whole lot of issues. That one thing seemed to improve performance by about 2-3x.
- Made validation testing in other places more efficient.

### DIFF
--- a/.github/workflows/server.yaml
+++ b/.github/workflows/server.yaml
@@ -45,7 +45,7 @@ jobs:
       with:
         path: ~/.cache/pip
         # Makes the hashing depend on changes to requirements.txt and python version
-        key: ${{ hashFiles('server/requirements.txt') }} py3.9.5
+        key: ${{ runner.os }}-${{ hashFiles('server/requirements.txt') }}-py3.9.5
     
     - name: Set up Python 3.9.5
       uses: actions/setup-python@v2
--- a/.gitignore
+++ b/.gitignore
@@ -339,6 +339,9 @@ dist
 # TODO: where does this rule come from?
 docs/_book
 
+# For local testing
+client/vue.config.js
+
 # TODO: where does this rule come from?
 test/
 
--- a/client/src/components/DoubleField.vue
+++ b/client/src/components/DoubleField.vue
@@ -34,22 +34,21 @@ export default {
   data() {
     return {
       current: this.init,
+      isValid: true,
     };
   },
   computed: {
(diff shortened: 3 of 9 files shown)


steven-murray (approved):
I can't say I understand all the code to change things, but the effect looks perfect! Thank you @aneuhold

PR #179, opened 2022-02-26
Timeouts and Dependencies Update

aneuhold:
### Overview

Quick update to timeouts in case the server takes a while to deploy.

There was also an issue with the older version of `astropy` which could no longer get `soft_unicode` from `markup_safe`. See [this Github post](https://github.com/aws/aws-sam-cli/issues/3661). The solution seemed to be to upgrade astropy to 4.2. This was tested locally and seems to work well, but am not sure about the implications of the upgrade on the theoretical aspects of the application.

Looking to Steven's idea on if this upgrade is okay, referencing [this changelog from astropy on the differences from 4.1 -> 4.2](https://docs.astropy.org/en/stable/changelog.html#id101). Again, this technically looks to be working, but not sure if the upgrade will have mathematical side-effects.

### Includes:
- `.github/workflows/serverDeployment.yaml`
- client/package.json 
- client/yarn.lock 
- server/docker-compose.yml 
- server/requirements.txt 

### Developer Checklist:

### DIFF
--- a/.github/workflows/clientDeployment.yaml
+++ b/.github/workflows/clientDeployment.yaml
@@ -60,7 +60,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
     
     - name: Install dependencies
-      run: cd client && npm install
+      run: cd client && yarn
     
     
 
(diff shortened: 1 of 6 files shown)


steven-murray (approved):
Thank you so much @aneuhold! The upgraded astropy looks to be fine.

